PR #43, opened 2017-01-06
Documentation typo: checkCancelled  -> checkCanceled

smarter:
Both are correct in English, but the compiler does not know that ;)


spoenemann:
Thanks!

PR #61, opened 2017-02-01
LSP version 3.0

akosyakov:


akosyakov:
There are 2 breaking changes:
- [jsonrpc] error code is not enum anymore but integer to support user defined error codes
- [lsp] a type of ServerCapabilities.textDocumentSync is Either now instead of Integer, but `JsonRpcData` active annotation infers setters for left and right type arguments of Either type, so at least it should not break on the server side

akosyakov:
There is one vague point for me: https://github.com/Microsoft/language-server-protocol/issues/165.

akosyakov:
@svenefftinge There is a new contract for `initialized` method: https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#initialize

> if the server receives request or notification before the initialize request it should act as follows:
> - for a request the respond should be errored with code: -32002. The message can be picked by the server.
> - notifications should be dropped.

> During the initialize request the server is allowed to sent the notifications window/showMessage, window/logMessage and telemetry/event as well as the window/showMessageRequest request to the client.

We can provide decorators for LanguageServer and LanguageClient which enforce the contract, but it still will be up to the client whether apply decorators or enforce the contract themself.

akosyakov:
A review of correspondence to the spec is welcome.

svenefftinge:
The initialization contract is something the actual implementers should take care of.

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/messages/ResponseError.java
```diff
@@ -16,16 +16,20 @@
...
-	private ResponseErrorCode code;
+	private int code;
 
-	public ResponseErrorCode getCode() {
+	public int getCode() {
 		return this.code;
 	}
 
-	public void setCode(ResponseErrorCode code) {
+	public void setCode(int code) {
```
svenefftinge: Why is this now an int?

akosyakov: @svenefftinge The protocol says that for requests custom error codes can be defined and the version 3.0 there is such for the initialization request.

Issue #71, opened 2017-02-21
Incompatibility with vscode-languageclient 3.0


smarter:
Using https://github.com/Microsoft/vscode-languageserver-node/tree/master/client 3.0.4:
```java
févr. 21, 2017 3:32:04 PM org.eclipse.lsp4j.jsonrpc.RemoteEndpoint handleNotification
AVERTISSEMENT: Notification threw an exception: {
  "jsonrpc": "2.0",
  "method": "initialized",
  "params": {}
}
java.lang.IllegalArgumentException: wrong number of arguments
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.lambda$null$0(GenericEndpoint.java:46)
        at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.notify(GenericEndpoint.java:87)
        at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.handleNotification(RemoteEndpoint.java:165)
        at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.consume(RemoteEndpoint.java:136)
        at org.eclipse.lsp4j.jsonrpc.Launcher.lambda$null$0(Launcher.java:59)
        at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.handleMessage(StreamMessageProducer.java:149)
        at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.listen(StreamMessageProducer.java:77)
        at org.eclipse.lsp4j.jsonrpc.json.ConcurrentMessageProcessor.run(ConcurrentMessageProcessor.java:68)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
```

[...]

PR #72, opened 2017-02-26
Fixes eclipse/lsp4j#71: ignore unexpected params

akosyakov:


org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/GenericEndpoint.java
```diff
@@ -28,6 +29,7 @@
 public class GenericEndpoint implements Endpoint {
 	
 	private static final Logger LOG = Logger.getLogger(GenericEndpoint.class.getName());
+	private static Object[] NO_ARGUMENTS = {};
```
kittaakos: It is a minor thing, and `NO_ARGUMENTS` an array anyway, but still, could we make this final?

akosyakov: @kittaakos thx!

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/GenericEndpoint.java
```diff
@@ -42,8 +44,9 @@ protected void recursiveFindRpcMethods(Object current, Set<Class<?>> visited, Se
 			@SuppressWarnings("unchecked")
 			Function<Object, CompletableFuture<Object>> handler = (arg) -> {
 				try {
-					Object[] argument = arg == null ? new Object[0] : new Object[] { arg };
-					return (CompletableFuture<Object>) methodInfo.method.invoke(current, argument);
+					Method method = methodInfo.method;
+					Object[] arguments = this.getArguments(method, arg);
```
kittaakos: Too much TS code, right? :) You could omit `this` just to align your changes with the existing code style.

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/GenericEndpoint.java
```diff
@@ -62,6 +65,17 @@ protected void recursiveFindRpcMethods(Object current, Set<Class<?>> visited, Se
 			}
 		});
 	}
+	
+	protected Object[] getArguments(Method method, Object arg) {
+		if (arg == null) {
+			return NO_ARGUMENTS;
+		}
+		if (method.getParameterCount() == 0) {
+			LOG.warning("Unexpected params '" + arg + "' for '" + method + "' is ignored");
```
kittaakos: Yet another small comment: here, maybe we could distinguish between single objects and Arrays to have a better log message.

akosyakov: actually the value does not matter, the point that the argument should not passed at all

PR #79, opened 2017-04-21
Ak/prepare for 0.2.0

akosyakov:


akosyakov:
Please review.

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/MessageTypeAdapterFactory.java
```diff
@@ -88,16 +88,20 @@ public Message read(JsonReader in) throws IOException {
...
+					if (parameterType == null || parameterType == Void.class) {
 						params = new JsonParser().parse(in);
-					else
-						params = gson.fromJson(in, type);
+					} else {
+						params = gson.fromJson(in, parameterType);
+						if (params instanceof JsonElement) {
+							params = gson.fromJson((JsonElement) params, parameterType);
+						}
+					}
```
spoenemann: The reason I had deferred the parsing of JsonElement to `createMessage(..)` is that I'd like to support an arbitrary order of message attributes. This is relevant when `params` appears before `method`. I vote for restoring the previous behavior, except for the special handling of `Void.class`.

akosyakov: ok, will do

akosyakov:
@spoenemann I've put deferred params parsing back and added tests for it

PR #90, opened 2017-05-04
Multiple parameters support

akosyakov:
Includes https://github.com/eclipse/lsp4j/pull/89

### DIFF
--- a/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/JsonRpcMethod.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/JsonRpcMethod.java
@@ -15,13 +15,15 @@ import java.lang.reflect.Type;
 public class JsonRpcMethod {
 	
 	private final String methodName;
-	private final Type parameterType;
+	private final Type[] parameterTypes;
 	private final Type returnType;
 	private final boolean isNotification;
 	
-	private JsonRpcMethod(String methodName, Type parameterType, Type returnType, boolean isNotification) {
+	private JsonRpcMethod(String methodName, Type[] parameterTypes, Type returnType, boolean isNotification) {
+		if (methodName == null)
+			throw new NullPointerException("methodName");
 		this.methodName = methodName;
-		this.parameterType = parameterType;
+		this.parameterTypes = parameterTypes;
 		this.returnType = returnType;
 		this.isNotification = isNotification;
 	}
@@ -30,8 +32,8 @@ public class JsonRpcMethod {
 		return methodName;
 	}
 	
-	public Type getParameterType() {
-		return parameterType;
+	public Type[] getParameterTypes() {
+		return parameterTypes;
 	}
 	
 	public Type getReturnType() {
@@ -41,17 +43,13 @@ public class JsonRpcMethod {
 	public boolean isNotification() {
 		return isNotification;
 	}
-
-	public static JsonRpcMethod notification(String name, Type parameterType) {
-		if (name == null)
-			throw new NullPointerException("name");
-		return new JsonRpcMethod(name, parameterType, Void.class, true);
+	
+	public static JsonRpcMethod notification(String name, Type... parameterTypes) {
+		return new JsonRpcMethod(name, parameterTypes, Void.class, true);
 	}
 	
-	public static JsonRpcMethod request(String name, Type parameterType, Type returnType) {
-		if (name == null)
-			throw new NullPointerException("name");
-		return new JsonRpcMethod(name, parameterType, returnType, false);
+	public static JsonRpcMethod request(String name, Type returnType, Type... parameterTypes) {
+		return new JsonRpcMethod(name, parameterTypes, returnType, false);
 	}
 	
 	@Override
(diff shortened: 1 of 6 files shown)


org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/JsonRpcMethod.java
```diff
@@ -62,26 +60,24 @@ public String toString() {
...
 	}
 
-	@Override
 	public int hashCode() {
 		final int prime = 31;
 		int result = 1;
 		result = prime * result + (isNotification ? 1231 : 1237);
 		result = prime * result + ((methodName == null) ? 0 : methodName.hashCode());
-		result = prime * result + ((parameterType == null) ? 0 : parameterType.hashCode());
+		result = prime * result + ((parameterTypes == null) ? 0 : parameterTypes.hashCode());
```
spoenemann: Would it make sense to use `Arrays.HashCode(..)` here?

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/JsonRpcMethod.java
```diff
@@ -97,10 +93,10 @@ public boolean equals(Object obj) {
 				return false;
 		} else if (!methodName.equals(other.methodName))
 			return false;
-		if (parameterType == null) {
-			if (other.parameterType != null)
+		if (parameterTypes == null) {
+			if (other.parameterTypes != null)
 				return false;
-		} else if (!parameterType.equals(other.parameterType))
+		} else if (!parameterTypes.equals(other.parameterTypes))
```
spoenemann: Would it make sense to use `Arrays.equals(..)` here?

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/MessageTypeAdapterFactory.java
```diff
@@ -122,24 +125,62 @@ public Message read(JsonReader in) throws IOException {
...
+		
+		protected Object fromJson(JsonElement element, Type type) {
+			if (isNull(element)) {
+				return null;
+			}
+			if (isNullOrVoidType(type)) {
+				return element;
+			}
+			Object value = gson.fromJson(element, type);
+			if (isNull(value)) {
```
spoenemann: The result of `gson.fromJson(element, type)` will never be a JsonNull?

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/MessageTypeAdapterFactory.java
```diff
@@ -122,24 +125,62 @@ public Message read(JsonReader in) throws IOException {
...
+			if (isNull(rawParams)) {
+				return null;
 			}
-			return null;
+			Type[] parameterTypes = getParameterTypes(method);
+			if (parameterTypes.length == 1) {
+				return fromJson(rawParams, parameterTypes[0]);
+			}
+			if (parameterTypes.length > 1 && rawParams instanceof JsonArray) {
+				List<Object> parameters = new ArrayList<Object>();
```
spoenemann: You could initialize the ArrayList with the size of the JsonArray.

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/MessageTypeAdapterFactory.java
```diff
@@ -88,12 +96,7 @@ public Message read(JsonReader in) throws IOException {
 					break;
 				}
 				case "params": {
-					Type parameterType = getParameterType(method);
-					if (isNullOrVoidType(parameterType)) {
-						params = new JsonParser().parse(in);
-					} else {
-						params = gson.fromJson(in, parameterType);
-					}
+					rawParams = new JsonParser().parse(in);
```
spoenemann: With this change we always create a JsonElement tree, which is less efficient than creating the parameters directly from the stream. Can we still support both without too much code overhead?

akosyakov:
@spoenemann i've addressed your comments, please review again.

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/MessageTypeAdapterFactory.java
```diff
@@ -128,11 +128,47 @@ public Message read(JsonReader in) throws IOException {
...
+					Type parameterType = index < parameterTypes.length ? parameterTypes[index] : null;
+					Object parameter = fromJson(in, parameterType);
+					parameters.add(parameter);
+					index++;
+				}
+				while (index < parameterTypes.length) {
+					parameters.add(null);
+					index++;
+				}
+				if (in.peek() == JsonToken.END_ARRAY) {
```
spoenemann: I don't think this check is necessary. Same for `in.peek() != JsonToken.END_ARRAY` in l. 146. The JsonReader should guarantee that the next token is `END_ARRAY` if and only if `in.hasNext()` is false.

spoenemann:
Found a minor issue, otherwise good work! 👍

PR #110, opened 2017-07-06
Additional properties of Formatting options must not be wrapped in a 'properties' property

### LINKED ISSUE #99: FormattingOptions model object
I have a question about implementation "Signature for further properties" of the FormattingOptions (https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#document-formatting-request) . lsp4j java class FormattingOptions (https://github.com/eclipse/lsp4j/blob/master/org.eclipse.lsp4j/src/main/xtend-gen/org/eclipse/lsp4j/FormattingOptions.java#L25) to represents this model description contains field "properties" like a map. Is it correct? I'm working on investigation using Clangd  language server and I see that lsp4j always send json with FormattingOptions which include field "properties":
 `{"jsonrpc":"2.0","id":"2","method":"textDocument/rangeFormatting","params":{"range":{"start":{"line":1,"character":4},"end":{"line":1,"character":12}},"textDocument":{"uri":"file:///home/antey/projects/cSimple/file.c"},"options":{"tabSize":4,"insertSpaces":true,"properties":{}}}}`
Clangd doesn't send response message for such request message. But if I delete "properties" field -Clangd successfully send me response:
` {"jsonrpc":"2.0","id":"2","method":"textDocument/rangeFormatting","params":{"range":{"start":{"line":1,"character":4},"end":{"line":1,"character":12}},"textDocument":{"uri":"file:///home/antey/projects/cSimple/file.c"},"options":{"tabSize":4,"insertSpaces":true}}}`
I try to understand where is a problem? Inside lsp4j library, or this is a bug Clangd, or root of this problem is unclear language server specification for FormattingOptions?

spoenemann:
Fixes #99. I tried to keep the protocol API as compatible as possible with the previous state.

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -1289,31 +1289,109 @@ class FileEvent {
 /**
  * Value-object describing what options formatting should use.
  */
-@JsonRpcData
-class FormattingOptions {
-	/**
-	 * Size of a tab in spaces.
-	 */
-	int tabSize
+class FormattingOptions extends LinkedHashMap<String, Object> {
 
-	/**
-	 * Prefer spaces over tabs.
-	 */
-	boolean insertSpaces
+	static val TAB_SIZE = 'tabSize'
+	static val INSERT_SPACES = 'insertSpaces'
 
-	/**
-	 * Signature for further properties.
-	 */
-	Map<String, String> properties
-    
     new() {
     }
     
-    new(int tabSize, boolean insertSpaces, Map<String, String> properties) {
+    new(int tabSize, boolean insertSpaces) {
     	this.tabSize = tabSize
     	this.insertSpaces = insertSpaces
-    	this.properties = properties
     }
+    
+    /**
+     * @deprecated See https://github.com/eclipse/lsp4j/issues/99
+     */
+    @Deprecated
+    new(int tabSize, boolean insertSpaces, Map<String, String> properties) {
+    	this(tabSize, insertSpaces)
+    	putAll(properties)
+    }
+    
+    /**
+     * Only {@code boolean | number | string} are accepted by formatting options.
+     */
+	override put(String key, Object value) {
+		switch key {
+			case TAB_SIZE: {
+				if (value instanceof Integer)
+		    		return super.put(key, value)
+		    	else if (value instanceof Number)
+		    		return super.put(key, value.intValue)
+		    	else if (value instanceof String)
+		    		try {
+		    			return super.put(key, Integer.valueOf(value))
+					} catch (NumberFormatException e) {}
+			}
+			case INSERT_SPACES: {
+				if (value instanceof Boolean)
+		    		return super.put(key, value)
+		    	else if (value instanceof String)
+		    		return super.put(key, Boolean.valueOf(value))
+			}
+			default: {
+				if (value instanceof Boolean || value instanceof Number || value instanceof String)
+					return super.put(key, value)
+				else
+					return super.put(key, value.toString)
+			}
+		}
+		return null
+	}
+    
+	/**
+	 * Size of a tab in spaces.
+	 */
+    def int getTabSize() {
+    	val value = get(TAB_SIZE)
+    	if (value instanceof Number)
+    		return value.intValue
+    	else
+    		throw new AssertionError("Property '" + TAB_SIZE + "' must be a number")
+    }
+    
+    def void setTabSize(int tabSize) {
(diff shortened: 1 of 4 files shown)


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -1289,31 +1289,109 @@ class FileEvent {
...
-    new(int tabSize, boolean insertSpaces, Map<String, String> properties) {
-    	this.tabSize = tabSize
-    	this.insertSpaces = insertSpaces
-    	this.properties = properties
+    /**
+     * @deprecated See https://github.com/eclipse/lsp4j/issues/99
+     */
+    @Deprecated
+    def Map<String, String> getProperties() {
+    	val result = newLinkedHashMap
```
szarnekow: This will silently break clients that did getProperties().put(..). Was that a valid use case? Maybe return a forwarding map instead?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -1289,31 +1289,109 @@ class FileEvent {
...
 	 */
-	int tabSize
-
-	/**
+    def int getTabSize() {
+    	val value = get(TAB_SIZE)
+    	if (value instanceof Number)
+    		return value.intValue
+    	else
+    		throw new AssertionError("Property '" + TAB_SIZE + "' must be a number")
```
szarnekow: This will be thrown for new instances that were created by the default ctor. In that case, no value is present (null) and null is not an instance of Number. Maybe return 0 as before?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -1289,31 +1289,109 @@ class FileEvent {
...
-	boolean insertSpaces
-
-	/**
-	 * Signature for further properties.
-	 */
-	Map<String, String> properties
+    def boolean isInsertSpaces() {
+       	val value = get(INSERT_SPACES)
+    	if (value instanceof Boolean)
+    		return value
```
szarnekow: Same here. Return false instead?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
spoenemann: Yes, forgot the null case. For other instances the AssertionError should still be valid.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
spoenemann: I'll replace it by `mapValues[toString]`, then it's no longer silent :-)

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -1289,31 +1289,109 @@ class FileEvent {
 /**
  * Value-object describing what options formatting should use.
  */
-@JsonRpcData
-class FormattingOptions {
+class FormattingOptions extends LinkedHashMap<String, Object> {
```
spoenemann: Maybe, but I don't like `Either<Boolean, Either<Number, String>>`. Should we add an `Either3` class so we can write `Either3<Boolean, Number, String>`?

akosyakov: We could, but we have already everything in place for `Either` as support in active annotations and JSON parsing.

AndrienkoAleksandr: +1@akosyakov

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -1289,31 +1289,112 @@ class FileEvent {
 /**
  * Value-object describing what options formatting should use.
  */
-@JsonRpcData
-class FormattingOptions {
+class FormattingOptions extends LinkedHashMap<String, Either<String, Either<Number, Boolean>>> {
```
svenefftinge: Number should be Integer

akosyakov: but in the protocol is js number, for custom properties float, double, etc. should be supported

PR #112, opened 2017-07-11
EndpointProxy: Fixed 'equals'

### LINKED ISSUE #106: `equals` seems broken for RemoteEndpoint
Here is a piece of code
```java
Collection<LanguageServer> file2LanguageServers = LanguageServiceAccessor.getLanguageServers(testFile2, c -> Boolean.TRUE); // 
LanguageServer ls = file2LanguageServers.iterator().next();
assertTrue(file2LanguageServers.contains(ls)); // FAILS, so I investigated with next lines
assertSame(ls, ls); // ok
assertEquals(ls, ls); // FAILS
```
`ls` is an instance of `org.eclipse.lsp4j.jsonrpc.RemoteEndpoint`.
`hashCode()` seems to be properly working as I can add the same `ls` multiple time in an `HashSet` without the set being modified. I imagine this is because the `HashSet` also checks for identity equality `==` before trying `equals()` so it's not facing the bug in `equals()`.
This is org.eclipse.lsp4j.jsonrpc_0.2.0.v20170518-0647.jar

spoenemann:
Fixes #106. The problem was that the proxy was calling `equals` on the delegate, which always returns false if the argument is the proxy. According to the contract of `equals`, we should return `true`.

### DIFF
--- a/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/EndpointProxy.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/EndpointProxy.java
@@ -9,6 +9,7 @@ package org.eclipse.lsp4j.jsonrpc.services;
 
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/EndpointProxy.java
```diff
@@ -68,6 +75,10 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
 		if (delegateInfo != null) {
 			return delegateInfo.delegate;
 		}
+		if (object_equals.equals(method) && args.length == 1 && args[0] == proxy) {
```
szarnekow: Maybe delegate hashCode and equals on the proxy to the EndpointProxy itself? Proxy.getInvocationHandler will allow you to obtain the InvokationHandler of the proxy argument that is passed into invoke. This would be more symmetric, I suppose. Also a proper toString could be implemented that way.

PR #116, opened 2017-08-22
Added Either3 type

spoenemann:
* Added a type `Either3<T1, T2, T3> extends Either<T1, Either<T2, T3>>`
 * FormattingOptions use Either3 as map values (see #110)
 * Added `TypeUtils` to extract common code from Gson type adapter factories
 * Corrected handling of generic types for Collection and Either (both type adapters were wrong in some cases)

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -1290,7 +1290,7 @@ class FileEvent {
 /**
  * Value-object describing what options formatting should use.
  */
-class FormattingOptions extends LinkedHashMap<String, Either3<String, Integer, Boolean>> {
+class FormattingOptions extends LinkedHashMap<String, Either3<String, Number, Boolean>> {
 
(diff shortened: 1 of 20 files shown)


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -1289,7 +1290,7 @@ class FileEvent {
 /**
  * Value-object describing what options formatting should use.
  */
-class FormattingOptions extends LinkedHashMap<String, Either<String, Either<Integer, Boolean>>> {
+class FormattingOptions extends LinkedHashMap<String, Either3<String, Integer, Boolean>> {
```
akosyakov: I thought `Number` should be used instead of `Integer`? I think the spec allows all kind of numbers. But it is a separate issue.

spoenemann: I changed that to `Integer` a while ago because of [a review comment](https://github.com/eclipse/lsp4j/pull/110#discussion_r129280523) by @svenefftinge, but I agree that Number should be correct here.

PR #117, opened 2017-09-04
Moved some methods from Either to TypeUtils

spoenemann:
`Either` contains some methods that are used only in its type adapter. They should be moved there to separate the data structure from its parsing / serialization code.


akosyakov:
I am not sure about making them protected and moving to the gson adapter. These methods constitute the reflection API for Either, they don't not depend on gson and can be used outside of json serialization/deserialization context.

spoenemann:
@akosyakov I moved the methods to TypeUtils (and shortened / renamed them).

PR #120, opened 2017-09-12
Send full stack trace to remote endpoint on internal error

### LINKED ISSUE #119: Provide exception details with ResponseError
Currently, we do not provide any information about the exception that caused a ResponseError.
In RemoteProxy we should `setData` to set the string representation of the throwable.

spoenemann:
Fixes #119.

### DIFF
--- a/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/RemoteEndpoint.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/RemoteEndpoint.java
@@ -7,6 +7,8 @@
  *******************************************************************************/
 package org.eclipse.lsp4j.jsonrpc;
 
+import java.io.ByteArrayOutputStream;
+import java.io.PrintWriter;
 import java.lang.reflect.InvocationTargetException;
 import java.util.LinkedHashMap;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/RemoteEndpoint.java
```diff
@@ -47,8 +49,13 @@
 		} else {
 			LOG.log(Level.SEVERE, "Internal error: " + throwable.getMessage(), throwable);
 			ResponseError error = new ResponseError();
-			error.setMessage("Internal error, please look at the server's logs.");
+			error.setMessage("Internal error.");
```
akosyakov: probably `throwable.getMessage()` should be also part of a message

akosyakov:
I wonder whether it is always the good thing to expose an internal exception information to the client. In VS code context it should not harm, not sure about servers running remotely.

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/RemoteEndpoint.java
spoenemann: The first line of the stack trace includes the message.

spoenemann:
In case you want to reduce the amount of information sent to the client, you can create the RemoteEndpoint with a different exception handler. My change only affects the default exception handler.

spoenemann:
I'll merge this so I can prepare release 0.3.0

Issue #128, opened 2017-11-13
IntegrationTest.testCancellation is intermittent


jonahkichwacoders:
I have run into testCancellation being intermittent on occasion, and can see that on [Jenkins](http://services.typefox.io/open-source/jenkins/job/lsp4j/job/master/75/testReport/org.eclipse.lsp4j.jsonrpc.test/IntegrationTest/testCancellation/) it occasionally fails in the same way:
```

java.lang.AssertionError: Timeout waiting for confirmation of cancellation.
	at org.junit.Assert.fail(Assert.java:88)
	at org.eclipse.lsp4j.jsonrpc.test.IntegrationTest.testCancellation(IntegrationTest.java:146)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
[...]

PR #129, opened 2017-11-13
Bug #128: Ensure that client is computing before issuing cancel

jonahkichwacoders:
This test is trying to verify that the cancel token was working
as expected. However there was a race condition in the test that
meant if the cancel was acted on before the computeAsync was run
the computeAsync would not be run. This caused an intermittent
test failure because cancellationHappened[0] would never be set
to true, even though the client would act on the cancellation.

Signed-off-by: Jonah Graham <jonah@kichwacoders.com>


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=128

spoenemann:
Thanks!

Issue #130, opened 2017-11-13
Support Debug Server Protocol


jonahkichwacoders:
The Debug Server Protocol, originating in VS code, provides a generic debug protocol with numerous "adapters" to connect VS code to different debuggers (such as Chrome debugger, gdb, etc.).

To support the Debug Server Protocol in Eclipse and Che, or to be able to write an adapter in Java, it will be useful to have a common library. This issue is to track extending the lsp4j library to support the debug protocol.

The protocol itself is defined at https://github.com/Microsoft/vscode-debugadapter-node/blob/master/debugProtocol.json

jonahkichwacoders:
PR #131 created.

mickaelistria:
Just out of curiosity: how easier it would have been to implement it if the debug protocol were using JSON-RCP? I'm asking this to evaluate whether it would still be interesting to influence the debug protocol in adopting JSON-RCP or to write a some JSON-RCP proxy for this Debug Server Protocol.

jonahkichwacoders:
It would certainly be easier. How much is harder to say, certainly days of work. I initially attacked everything in parallel, from the low level implementation of the protocol to the very complicated (but complete and extensible Eclipse Platform Debug)  to the simpler debug layer in Eclipse Che.

However, if I did not have lsp4j.jsonrpc to build on top of, it would have been even more work. Fortunately the two protocols are not that dissimilar, and now the work is done.

jonahkichwacoders:
This PR is a dependency of the Eclipse implementation (DSP4E, part of the LSP4E project) see: https://bugs.eclipse.org/bugs/show_bug.cgi?id=508470

jonahkichwacoders:
This has been merged now. There is more than likely going to be refinements, but they will best be handled in new bug reports (like #139)

mickaelistria:
Awesome!

PR #131, opened 2017-11-13
Bug #130: Support for JSON-RPC-like Debug Server Protocol's RPC

jonahkichwacoders:
This provides the supporting RPC implementation for VSCode's Debug Protocol
as defined https://github.com/Microsoft/vscode-debugadapter-node/blob/master/debugProtocol.json

See the class comment on DebugMessageTypeAdapterFactory for details on
mapping between Debug Server Protocol and JSON-RPC 2.0.

Signed-off-by: Jonah Graham <jonah@kichwacoders.com>

### DIFF
--- a/ide/build.gradle
+++ b/ide/build.gradle
@@ -1,3 +1,11 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2017 TypeFox GmbH (http://www.typefox.io) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+
 plugins {
 	id 'com.diffplug.gradle.oomph.ide' version '3.1.0'
 }
@@ -28,8 +36,10 @@ oomphIde {
 	// explicitly adding the folders is an easy workaround.
 	addProjectFolder('../')
 	addProjectFolder('../org.eclipse.lsp4j')
+	addProjectFolder('../org.eclipse.lsp4j.debug')
 	addProjectFolder('../org.eclipse.lsp4j.generator')
 	addProjectFolder('../org.eclipse.lsp4j.jsonrpc')
+	addProjectFolder('../org.eclipse.lsp4j.jsonrpc.debug')
 
 	// setup theming
 	style {
--- a/org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/DebugProtocol.xtend
+++ b/org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/DebugProtocol.xtend
@@ -1,0 +1,2470 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Kichwa Coders Ltd. and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+
+package org.eclipse.lsp4j.debug;
+
+import com.google.gson.annotations.SerializedName
+import java.util.Map
+import org.eclipse.lsp4j.generator.JsonRpcData
+import org.eclipse.lsp4j.jsonrpc.messages.Either
+
+/**
+ * Declaration of parameters, response bodies, and event bodies.
+ * <p>
+ * Auto-generated from debugProtocol.json schema version 1.24.0. Do not edit manually.
+ */
+class DebugProtcol {
+	/**
+	 * Version of debugProtocol.json this class was derived from.
+	 */
+	public static final String SCHEMA_VERSION = "1.24.0";
+}
+
+/**
+ * Event message for 'stopped' event type.
+ * <p>
+ * The event indicates that the execution of the debuggee has stopped due to some condition.
+ * <p>
+ * This can be caused by a break point previously set, a stepping action has completed, by executing a debugger
+ * statement etc.
+ */
+@JsonRpcData
+class StoppedEventArguments {
+	/**
+	 * The reason for the event.
+	 * <p>
+	 * For backward compatibility this string is shown in the UI if the 'description' attribute is missing (but it
+	 * must not be translated).
+	 * <p>
+	 * Possible values include - but not limited to those defined in {@link StoppedEventArgumentsReason}
+	 */
+	String reason;
+	/**
+	 * The full reason for the event, e.g. 'Paused on exception'. This string is shown in the UI as is.
+	 * <p>
+	 * This is an optional property.
+	 */
+	String description;
+	/**
+	 * The thread which was stopped.
+	 * <p>
+	 * This is an optional property.
+	 */
+	Integer threadId;
+	/**
+	 * Additional information. E.g. if reason is 'exception', text contains the exception name. This string is shown
+	 * in the UI.
+	 * <p>
+	 * This is an optional property.
+	 */
+	String text;
+	/**
+	 * If allThreadsStopped is true, a debug adapter can announce that all threads have stopped.
+	 * <p>
+	 *  The client should use this information to enable that all threads can be expanded to access their
+	 * stacktraces.
+	 * <p>
+	 *  If the attribute is missing or false, only the thread with the given threadId can be expanded.
+	 * <p>
+	 * This is an optional property.
+	 */
+	Boolean allThreadsStopped;
+}
+
+/**
+ * The reason for the event.
+ * <p>
+ * For backward compatibility this string is shown in the UI if the 'description' attribute is missing (but it
+ * must not be translated).
+ * <p>
+ * Possible values include - but not limited to those defined in {@link StoppedEventArgumentsReason}
+ */
+public interface StoppedEventArgumentsReason {
+	public static final String STEP = "step";
+	public static final String BREAKPOINT = "breakpoint";
+	public static final String EXCEPTION = "exception";
+	public static final String PAUSE = "pause";
+	public static final String ENTRY = "entry";
+}
+
+/**
+ * Event message for 'continued' event type.
+ * <p>
+ * The event indicates that the execution of the debuggee has continued.
+ * <p>
+ * Please note: a debug adapter is not expected to send this event in response to a request that implies that
+ * execution continues, e.g. 'launch' or 'continue'.
+ * <p>
+ * It is only necessary to send a ContinuedEvent if there was no previous request that implied this.
+ */
+@JsonRpcData
+class ContinuedEventArguments {
+	/**
+	 * The thread which was continued.
+	 */
+	Integer threadId;
+	/**
+	 * If allThreadsContinued is true, a debug adapter can announce that all threads have continued.
+	 * <p>
+	 * This is an optional property.
+	 */
+	Boolean allThreadsContinued;
+}
+
+/**
+ * Event message for 'exited' event type.
+ * <p>
+ * The event indicates that the debuggee has exited.
+ */
+@JsonRpcData
+class ExitedEventArguments {
+	/**
+	 * The exit code returned from the debuggee.
+	 */
+	Integer exitCode;
+}
+
+/**
+ * Event message for 'terminated' event types.
+ * <p>
+ * The event indicates that debugging of the debuggee has terminated.
+ */
+@JsonRpcData
+class TerminatedEventArguments {
+	/**
+	 * A debug adapter may set 'restart' to true (or to an arbitrary object) to request that the front end restarts
+	 * the session.
+	 * <p>
+	 * The value is not interpreted by the client and passed unmodified as an attribute '__restart' to the
+	 * launchRequest.
+	 * <p>
+	 * This is an optional property.
+	 */
+	Object restart;
+}
+
+/**
+ * Event message for 'thread' event type.
+ * <p>
+ * The event indicates that a thread has started or exited.
+ */
+@JsonRpcData
+class ThreadEventArguments {
+	/**
+	 * The reason for the event.
+	 * <p>
+	 * Possible values include - but not limited to those defined in {@link ThreadEventArgumentsReason}
+	 */
+	String reason;
+	/**
+	 * The identifier of the thread.
+	 */
+	Integer threadId;
+}
+
+/**
+ * The reason for the event.
+ * <p>
+ * Possible values include - but not limited to those defined in {@link ThreadEventArgumentsReason}
+ */
+public interface ThreadEventArgumentsReason {
+	public static final String STARTED = "started";
+	public static final String EXITED = "exited";
+}
+
+/**
+ * Event message for 'output' event type.
+ * <p>
+ * The event indicates that the target has produced some output.
+ */
+@JsonRpcData
+class OutputEventArguments {
+	/**
+	 * The output category. If not specified, 'console' is assumed.
+	 * <p>
+	 * This is an optional property.
+	 * <p>
+	 * Possible values include - but not limited to those defined in {@link OutputEventArgumentsCategory}
+	 */
+	String category;
+	/**
+	 * The output to report.
+	 */
+	String output;
+	/**
+	 * If an attribute 'variablesReference' exists and its value is > 0, the output contains objects which can be
+	 * retrieved by passing variablesReference to the VariablesRequest.
+	 * <p>
+	 * This is an optional property.
+	 */
+	Integer variablesReference;
+	/**
+	 * An optional source location where the output was produced.
+	 * <p>
+	 * This is an optional property.
+	 */
+	Source source;
+	/**
+	 * An optional source location line where the output was produced.
+	 * <p>
+	 * This is an optional property.
+	 */
+	Integer line;
+	/**
+	 * An optional source location column where the output was produced.
+	 * <p>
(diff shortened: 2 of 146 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=130

mickaelistria:
@svenefftinge @spoenemann @akosyakov Is there anything I can do to help on this topic? I'm eager to see progress there and can allocate most of my time.

svenefftinge:
@jonahkichwacoders thanks for the PR. 
Could we vote you in as a committer and maintainer of this part?

ide/build.gradle
```diff
@@ -30,6 +38,7 @@ oomphIde {
 	addProjectFolder('../org.eclipse.lsp4j')
 	addProjectFolder('../org.eclipse.lsp4j.generator')
 	addProjectFolder('../org.eclipse.lsp4j.jsonrpc')
+	addProjectFolder('../org.eclipse.dsp4j.jsonrpc')
```
spoenemann: I'm not sure about the namespace `org.eclipse.dsp4j`, since theoretically that could cause name clashes with other Eclipse projects. On the other hand, I like the symmetry with `org.eclipse.lsp4j`. What do others think about this?

An alternative project name could be `org.eclipse.lsp4j.jsonrpc.debug`

org.eclipse.dsp4j.jsonrpc/src/main/java/org/eclipse/dsp4j/jsonrpc/adapters/DebugMessageTypeAdapterFactory.java
```diff
@@ -0,0 +1,493 @@
...
+				}
+			}
+			if (type == null)
+				result = new JsonParser().parse(in);
+			else
+				result = gson.fromJson(in, type);
+			return result;
+		}
+
+		protected Object parseParams(JsonReader in, String method) throws IOException {
```
spoenemann: We should make `MessageTypeAdapterFactory` extensible in order to avoid duplicating code here. I'll prepare a PR for that.

org.eclipse.dsp4j.jsonrpc/src/main/java/org/eclipse/dsp4j/jsonrpc/json/DebugMessageJsonHandler.java
```diff
@@ -0,0 +1,38 @@
...
+public class DebugMessageJsonHandler extends MessageJsonHandler {
+	public DebugMessageJsonHandler(Map<String, JsonRpcMethod> supportedMethods) {
+		super(supportedMethods);
+	}
+
+	public DebugMessageJsonHandler(Map<String, JsonRpcMethod> supportedMethods, Consumer<GsonBuilder> configureGson) {
+		super(supportedMethods, configureGson);
+	}
+
+	public GsonBuilder getDefaultGsonBuilder() {
```
spoenemann: This additional class could be avoided by passing a `Consumer<GsonBuilder>` to the default `MessageJsonHandler`. When multiple type adapter factories are configured for the same type, the last one should win.

org.eclipse.dsp4j.jsonrpc/src/test/java/org/eclipse/dsp4j/jsonrpc/test/json/DebugMessageJsonHandlerTest.java
```diff
@@ -0,0 +1,582 @@
...
+			for (j = i; j <= n; j++) {
+				swap(properties, i, j);
+				testAllPermutations(properties, i + 1, n, test);
+				swap(properties, i, j);
+			}
+		}
+	}
+
+	@Test
+	public void testParamsParsing_AllOrders() {
```
spoenemann: This parsing test that tries all iterations would also be nice for the `MessageJsonHandlerTest` in `org.eclipse.lsp4j.jsonrpc`.

spoenemann:
@eclipsewebmaster 

> Issue tracker reference:
> https://bugs.eclipse.org/bugs/show_bug.cgi?id=130

That makes no sense. Has probably been triggered by Jonahs mention of `Bug #130`. So better use the term "issue" in this context :-D

jonahkichwacoders:
> Could we vote you in as a committer and maintainer of this part?
Yes.

jonahkichwacoders:
>Thank you for this proposal! The code seems to me of very high quality. I would support your nomination as a committer :-)

Thanks.

> Are you planning to also implement service interfaces with corresponding data structures in the style of org.eclipse.lsp4j.services.LanguageServer? Or have you already done that?

I am planning to create a PR in the coming days with the equivalent for the debug protocol. It is generated from the debug protocol json schema. Originally I was going to put it in a plug-in called org.eclipse.dsp4j, but based on your other comments, perhaps org.eclipse.lsp4j.debug would make more sense.

ide/build.gradle
jonahkichwacoders: OK, depending on decision here will affect name of the interfaces containing plug-in. I was originally planning on org.eclipse.dsp4j, but perhaps org.eclipse.lsp4j.debug makes more sense.

org.eclipse.dsp4j.jsonrpc/src/main/java/org/eclipse/dsp4j/jsonrpc/adapters/DebugMessageTypeAdapterFactory.java
jonahkichwacoders: OK, there is certainly some duplication between the two.

org.eclipse.dsp4j.jsonrpc/src/main/java/org/eclipse/dsp4j/jsonrpc/json/DebugMessageJsonHandler.java
jonahkichwacoders: OK, the advantage of your approach too is that the other type factories don't need to be duplicated. A lot of the tests instantiate `DebugMessageJsonHandler` directly, so I'll probably want a method somewhere that constructs a `MessageJsonHandler` correctly for the tests in the same way as it is constructed for by the `DebugLauncher`. Perhaps a method in the `DebugLauncher` itself.

org.eclipse.dsp4j.jsonrpc/src/test/java/org/eclipse/dsp4j/jsonrpc/test/json/DebugMessageJsonHandlerTest.java
jonahkichwacoders: OK, I will contribute it back to both places. I also realised on revisiting that the idea was to remove the _02/_03/etc versions of the methods, so I will do that too.

jonahkichwacoders:
> So better use the term "issue" in this context :-D

OK.

spoenemann:
It seems like the word "bug" is not necessary for triggering the Bugzilla bot, see eclipse/kapua#131

ide/build.gradle
svenefftinge: Yes, IIRC there even is some rule at eclipse for java packages to use the project name as prefix.

org.eclipse.dsp4j.jsonrpc/src/main/java/org/eclipse/dsp4j/jsonrpc/json/DebugMessageJsonHandler.java
jonahkichwacoders: I am having trouble writing a sensible method in DebugLauncher, at the moment it looks like the sample below, which seems uglier than having the extra class, ideas welcome.

The reason for the complication is the circular dependency between the `MessageJsonHandler` and the `DebugMessageTypeAdapterFactory`.

My next update will have the `DebugMessageJsonHandler` still, but rather than repeating everything in `getDefaultGsonBuilder` I'll call super's one and then add on the `DebugMessageTypeAdapterFactory`.

If you have a thought on something I overlooked, I would be pleased for the feedback.

```java

	static MessageJsonHandler createMessageHandler(Map<String, JsonRpcMethod> supportedMethods,
			Consumer<GsonBuilder> configureGson) {
		DebugMessageTypeAdapterFactory adapterFactory[] = new DebugMessageTypeAdapterFactory[] {null};
		Consumer<GsonBuilder> consumer = (builder) -> {
			builder.registerTypeAdapterFactory(adapterFactory[0]);
			configureGson.accept(builder);
		};
		MessageJsonHandler messageJsonHandler = new MessageJsonHandler(supportedMethods, consumer);
		adapterFactory[0] = new DebugMessageTypeAdapterFactory(messageJsonHandler);
		return messageJsonHandler;
	}
```

spoenemann: Hmmm, we could change the `configureGson` parameter in MessageJsonHandler to `BiConsumer<GsonBuilder, MessageJsonHandler>`, but that's a breaking change. @akosyakov, @svenefftinge do you think this is worth it? Would help for cases when someone wants to customize parsing / serialization of messages.

jonahkichwacoders: We could also add an additional constructor and deprecate the existing one. That would prevent it being a breaking change.

spoenemann: Right, we don't even need to deprecate the old one.

org.eclipse.dsp4j.jsonrpc/src/main/java/org/eclipse/dsp4j/jsonrpc/adapters/DebugMessageTypeAdapterFactory.java
spoenemann: Merged #135.

jonahkichwacoders: I updated the debug code to to use this new code. I also slightly refactored the `MessageTypeAdapterFactory` to access some of its logic from the subclass.

org.eclipse.dsp4j.jsonrpc/src/main/java/org/eclipse/dsp4j/jsonrpc/json/DebugMessageJsonHandler.java
jonahkichwacoders: I have not made this change yet. The current version, and the new constructor we discussed suffers from a problem. It passes a reference to `this` from the constructor when the object is not fully created, so creating new API that required and actually enforced passing around an incompletely constructed object seems like a bad idea. If you would like me to pursue this further (e.g. add an `initialize()` method for example) I can raise and issue and work on that separately?

org.eclipse.dsp4j.jsonrpc/src/test/java/org/eclipse/dsp4j/jsonrpc/test/json/DebugMessageJsonHandlerTest.java
jonahkichwacoders: I have added the `_AllOrders` method to `MessageJsonHandlerTest` and extended it to do all the types of messages.

jonahkichwacoders:
@spoenemann I have updated the code with all your feedback and it is again ready for review.

This update has the next part too, the service protocol files. As the DSP has a [Json Schema](https://github.com/jonahkichwacoders/vscode-debugadapter-node/blob/53e08384d5ab0786a053ae24acdcfa0a081fb95f/debugProtocol.json) that defines it, the `DebugProtocol.java`, `IDebugProtocolClient` and `IDebugProtocolServer` are auto-generated using a [generator](https://github.com/jonahkichwacoders/vscode-debugadapter-node/blob/53e08384d5ab0786a053ae24acdcfa0a081fb95f/src/generator.ts) derived from the [typescript generator](https://github.com/Microsoft/vscode-debugadapter-node/blob/99528cd3c2db13f3fd78047c5b6967bf13212c23/src/generator.ts) used by vscode. Please see the instructions for the generator in the root of the `org.eclipse.lsp4j.debug` plug-in. I don't know if the licensing of the generator allows it to live in the lsp4j code base. I also don't know if continuing to generate it automatically as the protocol evolves is the best plan either, so perhaps the generator is only useful to bootstrap the project?

org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/DebugProtocol.java
```diff
@@ -0,0 +1,11631 @@
...
+	 * <p>
+	 * The event indicates that the execution of the debuggee has stopped due to
+	 * some condition.
+	 * </p>
+	 * <p>
+	 * This can be caused by a break point previously set, a stepping action has
+	 * completed, by executing a debugger statement etc.
+	 * </p>
+	 */
+	public static class StoppedEventArguments {
```
spoenemann: The class structure should not be nested, but flat, in order to be consistent with the LSP implementation.

org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/DebugProtocol.java
```diff
@@ -0,0 +1,11631 @@
...
+		 * </p>
+		 * <p>
+		 * For backward compatibility this string is shown in the UI if the
+		 * 'description' attribute is missing (but it must not be translated).
+		 * </p>
+		 * <p>
+		 * Possible values include - but not limited to those defined in {@link Reason}
+		 * </p>
+		 */
+		public String reason;
```
spoenemann: The fields should not be public, but accessible via getters, in order to be consistent with the LSP implementation.

org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/services/IDebugProtocolClient.java
```diff
@@ -0,0 +1,218 @@
...
+	 * <li>frontend sends a SetExceptionBreakpointsRequest if one or more
+	 * exceptionBreakpointFilters have been defined (or if
+	 * supportsConfigurationDoneRequest is not defined or false)</li>
+	 * <li>frontend sends other future configuration requests</li>
+	 * <li>frontend sends one ConfigurationDoneRequest to indicate the end of the
+	 * configuration</li>
+	 * </ul>
+	 */
+	@JsonNotification
+	default void initialized() {
```
spoenemann: Why these default implementations? I'd remove them.

org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/DebugProtocol.java
```diff
@@ -0,0 +1,11631 @@
...
+ * <p>
+ * Auto-generated from debugProtocol.json schema version 1.24.0. Do not edit
+ * manually.
+ * </p>
+ */
+public class DebugProtocol {
+	/**
+	 * <p>
+	 * Version of debugProtocol.json this class was derived from.
+	 * </p>
```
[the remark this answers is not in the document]
jonahkichwacoders: I can add some additional logic to the generator to only add the `<p>` between paragraphs and omit them on single paragraph comments.

org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/DebugProtocol.java
jonahkichwacoders: Probably it would be better to generate a `DebugProtocol.xtend` with `@JsonRpcData` tags and that gets us even closer. It would be a 2 stage generation process (first the typescript one then the o.e.lsp4j.generator one), but then the `DebugProtocol.xtend` would be much readable (and shorter). WDYT?

spoenemann: Two-stage generation sounds a bit scary at first. But a nice side effect is that all the derived members (getters, setters, equals, hashCode) are consistent in both protocols. And I like the idea to have a more readable version of the Java data structures.

jonahkichwacoders:
@spoenemann Its ready, I have hopefully addressed all your concerns adequately. With the generation of `DebugProtocol.xtend` and using the existing generator as the second-pass generator makes it much easier to read `DebugProtocol.xtend` and the generated javadoc still looks fine.

org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/DebugProtocol.xtend
```diff
@@ -0,0 +1,2470 @@
...
+import java.util.Map
+import org.eclipse.lsp4j.generator.JsonRpcData
+import org.eclipse.lsp4j.jsonrpc.messages.Either
+
+/**
+ * Declaration of parameters, response bodies, and event bodies.
+ * <p>
+ * Auto-generated from debugProtocol.json schema version 1.24.0. Do not edit manually.
+ */
+class DebugProtcol {
```
spoenemann: I think this class could be removed now.

Issue #132, opened 2017-11-14
Support workspaceFolders proposal


mickaelistria:
https://github.com/Microsoft/language-server-protocol/issues/298#issuecomment-336700314 covers a discussion about allowing multi root/projects in language servers. This is not finalized in the protocol specification yet, but the vscode client api already covers it. This seems to be an important step forward and a major change, so it would be nice to start investigating support for it soon.

See https://bugs.eclipse.org/bugs/show_bug.cgi?id=526489 (PS: you should ask webmaster to remove bugzilla component if you're not using it).

martinlippert:
Is there a way to get the related pull request accepted and a snapshot build that includes this up somewhere? I am eagerly waiting for this to make our language servers ready for the latest VSCode 1.18 release (which released support for multi-root workspaces).

mickaelistria:
FYI, I've prepared a patch in LSP4E which takes advantage of this change in LSP4J to implement multi-root support, and it seems to work with JDT-LS https://git.eclipse.org/r/#/c/111750/
That doesn't validate everything though as JDT-LS ignores the "static" capabilities and mostly go through the registerCapability approach; but the operations to change workspaceFolder seem to work as expected between LSP4E and JDT-LS, using LSP4J as API on client side.

svenefftinge:
closing since https://github.com/eclipse/lsp4j/pull/133 is merged

PR #133, opened 2017-11-14
Issue #132: Proposal for workspaceFolders

mickaelistria:
Signed-off-by: Mickael Istria <mistria@redhat.com>

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -15,6 +15,7 @@ import org.eclipse.lsp4j.generator.JsonRpcData
 import org.eclipse.lsp4j.jsonrpc.messages.Either
 import org.eclipse.lsp4j.jsonrpc.messages.Either3
 import org.eclipse.lsp4j.jsonrpc.validation.NonNull
+import com.google.common.annotations.Beta
 
 @JsonRpcData
 class DynamicRegistrationCapabilities {
@@ -121,6 +122,13 @@ class WorkspaceClientCapabilities {
      * Capabilities specific to the `workspace/executeCommand` request.
      */
     ExecuteCommandCapabilities executeCommand
+    
+    /**
+     * Capabilities specific to the `workspace/didChangeWorkspaceFolders` notification.
+     * 
+     * This API is a <b>proposal</b> from LSP and may change.
+     */
+    @Beta Boolean workspaceFolders
 }
 
 @JsonRpcData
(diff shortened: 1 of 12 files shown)


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/LanguageClient.java
```diff
@@ -85,4 +87,14 @@
...
+
+	/**
+	 * The workspace/workspaceFolders request is sent from the server to the client
+	 * to fetch the current open list of workspace folders.
+	 * @return null in the response if only a single file is open in the tool,
+	 *         an empty array if a workspace is open but no folders are configured,
+	 *         the workspace folders otherwise.
+	 */
+	@JsonRequest("workspace/workspaceFolders")
+	CompletableFuture<List<WorkspaceFolder>> workspaceFolders();
```
spoenemann: Adding a method here breaks existing implementations. Since the protocol extension is still in a "proposed" state, we should avoid breaking changes. Either move the method to a separate interface or add a default implementation here. I prefer the second option.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/WorkspaceService.java
```diff
@@ -57,4 +58,14 @@
...
+
+	/**
+	 * The workspace/didChangeWorkspaceFolders notification is sent from the client
+	 * to the server to inform the server about workspace folder configuration changes.
+	 * The notification is sent by default if both ServerCapabilities/workspaceFolders
+	 * and ClientCapabilities/workapce/workspaceFolders are true; or if the server has
+	 * registered to receive this notification it first.
+	 */
+	@JsonNotification
+	void didChangeWorkspaceFolders(DidChangeWorkspaceFoldersParams params);
```
spoenemann: This breaks existing implementations, see above.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -1901,6 +1906,11 @@ class ServerCapabilities {
      * Experimental server capabilities.
      */
     Object experimental
+
+    /**
+     * Capabilities specific to the `workspace/didChangeWorkspaceFolders` notification.
+     */
+    WorkspaceFoldersOptions workspaceFolders
```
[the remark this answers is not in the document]
mickaelistria: I think that's more something to fix on the proposed extension side ;)

mickaelistria:
PR updated. Only the first comment has not been fixed since I'm waiting for a clarification on the proposal.

mickaelistria:
PR updated. Only the first comment has not been fixed since I'm waiting for a clarification on the proposal: https://github.com/Microsoft/language-server-protocol/issues/298#issuecomment-345271669

dbaeumer:
Commented in https://github.com/Microsoft/language-server-protocol/issues/298#issuecomment-345271669

mickaelistria:
Thanks for the clarification @dbaeumer .
I did update the pull request accordingly, introducing a WorkspaceServerCapabilities node in the model.

spoenemann:
👍 
I'll update the version to 0.4.0-SNAPSHOT tomorrow or so.

PR #135, opened 2017-11-20
Made Gson type adapters better extensible

spoenemann:
This will help reducing the amount of duplicated code in #131.


spoenemann:
The idea: Instead of
```
public class MyTypeAdapterFactory implements TypeAdapterFactory {
    ...
    private static class Adapter extends TypeAdapter<...> {
        ...
    }
}
```
we now have
```
public class MyTypeAdapter extends TypeAdapter<...> {
    public static class Factory implements TypeAdapterFactory {
        ...
    }
    ...
}
```

PR #154, opened 2018-02-01
Parse Object into a JsonElement

### LINKED ISSUE #126: "any" should unmarshal to JsonElement, not Object
In a couple of places in LSP, any json is allowed (for example here: https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#workspace_executeCommand)
Currently, the declared type of the parameters in WorkspaceService#executeCommand() is ```List<Object>```, and at runtime, the object is a structure made up of lists and maps.
I believe it would be easier for users if the declared type of "any" in lsp4j would be either JsonElement or String (the second passing the raw json to the implementor).

spoenemann:
Fixes #126.


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -444,6 +446,7 @@ class ClientCapabilities {
     /**
      * Experimental client capabilities.
      */
+    @JsonAdapter(JsonElementTypeAdapter.Factory)
     Object experimental
```
akosyakov: Why not use `JsonElement`? As far as I understood implementors need to downcast it always?

spoenemann: Because in this way implementors are still free to put arbitrary objects into those fields. If we would change the field type to JsonElement, implementors would have to take care of the serialization of their data (of course you are still free to put a JsonElement into the field).

akosyakov: ok

PR #155, opened 2018-02-01
Changed type of 'id' fields to Either<String, Number>

### LINKED ISSUE #138: JSON RPC specification violation on ID field.
Hi,

According to the JSON RPC 2.0 specification, the "id" field `It MUST be the same as the value of the id member in the Request Object.`. But if the request message send the 'id' value as an integer, it will get changed to a string[1] in the response.

[1] -
 https://github.com/eclipse/lsp4j/blob/master/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/messages/ResponseMessage.java#L25

Regards,
Hemika

spoenemann:
`id` fields in RequestMessage and ResponseMessage now accept both String and int values as demanded by [the specification](http://www.jsonrpc.org/specification). The id of every response is exactly the same value as given by the corresponding request, i.e. a String yields a String and an int yields an int.

Fixes #138.

### DIFF
--- a/org.eclipse.lsp4j.jsonrpc.debug/src/main/java/org/eclipse/lsp4j/jsonrpc/debug/DebugRemoteEndpoint.java
+++ b/org.eclipse.lsp4j.jsonrpc.debug/src/main/java/org/eclipse/lsp4j/jsonrpc/debug/DebugRemoteEndpoint.java
@@ -34,7 +34,7 @@ public class DebugRemoteEndpoint extends RemoteEndpoint {
 	@Override
 	protected DebugRequestMessage createRequestMessage(String method, Object parameter) {
 		DebugRequestMessage requestMessage = new DebugRequestMessage();
-		requestMessage.setId(String.valueOf(nextSeqId.incrementAndGet()));
+		requestMessage.setId(nextSeqId.incrementAndGet());
 		requestMessage.setMethod(method);
 		requestMessage.setParams(parameter);
 		return requestMessage;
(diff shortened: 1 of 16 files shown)


akosyakov:
@jonahkichwacoders could you review a debugging part, please?

org.eclipse.lsp4j.jsonrpc.debug/src/main/java/org/eclipse/lsp4j/jsonrpc/debug/DebugRemoteEndpoint.java
```diff
@@ -34,7 +34,7 @@ public DebugRemoteEndpoint(MessageConsumer out, Endpoint localEndpoint,
 	@Override
 	protected DebugRequestMessage createRequestMessage(String method, Object parameter) {
 		DebugRequestMessage requestMessage = new DebugRequestMessage();
-		requestMessage.setId(String.valueOf(nextSeqId.incrementAndGet()));
+		requestMessage.setId(nextSeqId.incrementAndGet());
```
jonahkichwacoders: This resolves one of the annoying issues I had, I really didn't like the converting back and forth, so thanks for this.

spoenemann:
> It seems fine to me in practice. But I wonder if:
> `MethodProvider.resolveMethod(String)` should be updated to be `MethodProvider.resolveMethod(Either<String, Number>)`?

Internally we still use String to store request ids. I think it's ok to assume that ids must be unique regardless of their type, i.e. a message with id `"2"` matches one with id `2`. This makes us robust against remote endpoints that arbitrarily switch between the two formats. The changes made by this PR are rather targeted to make picky remote endpoints happy: if the remote requested responses with integer ids, it will get integer ids.

PR #163, opened 2018-02-23
Add file level operations to WorkspaceEdit

gorkem:
Adds the new ResourceChange change type to WorkspaceEdit
also augments the WorkspaceEditCapabilities with the
new resourceChanges capability

See also: https://github.com/Microsoft/vscode-languageserver-node/pull/295
https://github.com/Microsoft/language-server-protocol/issues/272

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -41,6 +41,12 @@ class WorkspaceEditCapabilities {
 	 */
 	Boolean documentChanges
     
+  /**
+   * The client supports resource changes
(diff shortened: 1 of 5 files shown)


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -2315,6 +2321,36 @@ class TextDocumentEdit {
...
+ */
+@JsonRpcData
+class ResourceChange{
+
+  /**
+  * The Uri for current resource. Required for delete and move operations
+  * otherwise it is null.
+  *
+  */
+  @NonNull
```
spoenemann: The comment implies that this can be null, so the annotation should be removed.

gorkem:
Removed the NonNull annotation.

PR #164, opened 2018-02-27
Added a utility class for process handling

### LINKED ISSUE #161: Shutdown if Parent Process Shuts Down
LS does not exit if the parent process is not alive
see https://github.com/eclipse/xtext-core/issues/552

spoenemann:
Fixes #161 by offering a utility method that can be used by language server implementations to watch the parent process.


spoenemann:
After rethinking, we shouldn't offer a new API unless it works on all platforms. If anyone has an idea how it could be implemented for Windows, please share it.

PR #165, opened 2018-02-27
ClientCapabilities: Update with capabilities from latest 3.x protocol.

adityasharad:
The latest [LSP 3.x specification](https://microsoft.github.io/language-server-protocol/specification) adds a number of optional capabilities to `ClientCapabilities` and the types it references.
This PR adds support for those optional fields to the generated classes, creating classes to model new capabilities where required.

The additions (grouped by their parent field in `TextDocumentClientCapabilities`) are:
* `hover`: `contentFormat`
* `completionItem`: `commitCharactersSupport` and `documentationFormat`
* `completion`: `completionItemKind` and `contextSupport`
* `signatureHelp`: `signatureInformation`
* `documentSymbol`: `symbolKind`
* `typeDefinitions`
* `implementation`
* `colorProvider`


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/MarkupKind.java
```diff
@@ -0,0 +1,44 @@
...
+package org.eclipse.lsp4j;
+
+/**
+ * Describes the content type that a client supports in various
+ * result literals like `Hover`, `ParameterInfo` or `CompletionItem`.
+ *
+ * Please note that `MarkupKind`s must not start with a `$`. These kinds
+ * are reserved for internal usage.
+ */
+public enum MarkupKind {
```
spoenemann: Have you tested this? Our current `EnumTypeAdapter` assumes that enums are represented with numbers, not strings.

adityasharad:
@spoenemann thank you for the review. I will replace the `MarkupKind` enum with string constants, and add tests for sending client capabilities that exercise the new fields.

I did not intend to make a complete upgrade to 3.6 in this PR, instead focusing first on the `ClientCapabilities` changes. However, I do intend to add the 3.6 features that use the newly-introduced `MarkupKind`, for example the hover request, in later commits.

adityasharad:
@spoenemann I have added the tests you requested, and updated `TextDocumentClientCapabilities`.
I have not yet changed `WorkspaceClientCapabilities`: are you happy for me to drop the `Beta` label on the `workspaceFolders` field there, since it is now in 3.6?

Before I make use of `MarkupKind`/`MarkupContent` in messages: the `hover` message can now have a `MarkedString`, a list of `MarkedString`s, or a `MarkupContent` in its contents. This is currently represented as `List<Either<String, MarkedString>>`. How would you suggest this be changed to include `MarkupContent`?

spoenemann:
Thank you!

For protocol classes with a lot of properties we have omitted the constructors because we cannot provide constructors for all permutations of properties, see e.g. WorkspaceClientCapabilites. After this change this would also apply to CompletionCapabilities, especially because the constructors would become quite ambiguous with the two Boolean properties. I'd propose to omit the additional constructors for that class.

> are you happy for me to drop the Beta label on the workspaceFolders field there, since it is now in 3.6?

Yes.

> How would you suggest this be changed to include MarkupContent?

I fear the type declaration `MarkedString | MarkedString[] | MarkupContent` cannot be realized with our current implementation of `Either`. I need to think about that...

adityasharad:
Any thoughts so far on including `MarkupContent`? How about a 3-way `Either`?

spoenemann:
@adityasharad That doesn't work for the same reason as #173. I fear the only way to solve this is by implementing a special Gson TypeAdapter for the `Hover` class.

I propose you prepare this PR for merging and ignore the Hover problem for now:

 * Remove constructor as suggested [above](https://github.com/eclipse/lsp4j/pull/165#issuecomment-372251350)
 * Drop the Beta label on the workspaceFolders
 * Resolve conflict in test file
 * Add Signed-off-by commit footer

Then I'll merge it and implement the missing Hover TypeAdapter. Ok?

adityasharad:
Thanks @spoenemann, addressed your comments and squashed into one commit.

(Some of the conflicts were due to your recent changes to `JsonSerializeTest` being generated with Java 7 `Procedure1.apply` rather than Java 8 lambdas. I have resolved those in favour of the existing Java 7 code, but you may want to make the file consistently use either the Java 7 or 8 style.)

spoenemann:
👍

spoenemann:
See #175.

PR #174, opened 2018-04-05
SymbolKind: added more values from the language serveral protocol specification

joshtynjala:
As documented here: https://microsoft.github.io/language-server-protocol/specification#textDocument_documentSymbol


spoenemann:
Thank you!

PR #175, opened 2018-04-10
Support hover results

spoenemann:
* Added MarkupContent as type alternative to Hover contents.
 * Added a Gson type adapter for Hover.
 * Added a subclass of MessageJsonHandler for LSP that includes the HoverTypeAdapter.
 * Added a Builder class for Launcher in order to easily customize the JSON-RPC wiring.
 * Specialized the Launcher Builder for LSP and the debug protocol.

This change breaks the API of Hover results, but as far as I can tell there's no other way to support the [current specification of Hover results](https://microsoft.github.io/language-server-protocol/specification#hover-request-leftwards_arrow_with_hook).


spoenemann:
I added the Builder for Launcher so customizations such as additional type adapters will be easier in the future. However, there is a more general approach that we might consider:

 * Implement `Either4` in addition to the existing `Either3`.
 * Introduce an annotation `@EitherTypeSelect` that can be used to guide the EitherTypeAdapter when its alternatives cannot be chosen based on the next JSON token, but require a look-ahead of the object properties.
 * Use this annotation in Hover:
```xtend
@JsonRpcData
class Hover {
    @NonNull
    @EitherTypeSelect(type=MarkedString, propertyName='language')
    @EitherTypeSelect(type=MarkupContent, propertyName='MarkupKind')
    Either4<List<Either<String, MarkedString>>, String, MarkedString, MarkupContent> contents

    Range range
}
```

With this additional information the EitherTypeAdapter can parse a MarkedString when it finds a property named `language` and a MarkupContent when it finds a property named `MarkupKind`.

This approach would probably be quite over-engineered for just implementing Hover, but we could reconsider it the next time we get into a similar situation. What do you think?

spoenemann:
I simplified this change by using the JsonAdapter annotation instead of a custom MessageJsonHandler. This allowed me to split off the Launcher enhancements to #177.

spoenemann:
I realized that the [current version of VersionedTextDocumentIdentifier](https://microsoft.github.io/language-server-protocol/specification#versionedtextdocumentidentifier) also requires a special TypeAdapter :-/

According to the specification, the value of `version` must be either a number or `null`, but Gson allows to enable the serialization of null values only globally.

I was not very keen on writing yet another boilerplate TypeAdapter implementation, so I wrote an Active Annotation instead :-)

spoenemann:
This is weird:

```typescript
InitializeParams {
	/**
	 * The workspace folders configured in the client when the server starts.
	 * This property is only available if the client supports workspace folders.
	 * It can be `null` if the client supports workspace folders but none are
	 * configured.
	 *
	 * Since 3.6.0
	 */
	workspaceFolders?: WorkspaceFolder[] | null;
}
```

It means both `undefined` and `null` are possible and both have their own meaning. BAD!

Should we ignore this? I can't think of any sensible solution.

org.eclipse.lsp4j.generator/src/main/java/org/eclipse/lsp4j/generator/JsonRpcData.xtend
```diff
@@ -12,6 +12,10 @@ import java.lang.annotation.Retention
 import java.lang.annotation.Target
 import org.eclipse.xtend.lib.macro.Active
 
+/**
+ * Generates getters and setters for all fields as well as {@code eauals} and {@code hashCode} implementations.
```
akosyakov: `eauals` -> `equals`

spoenemann: I'll fix this in the next PR :-)

szarnekow:
> It means both undefined and null are possible and both have their own meaning. BAD!
Should we ignore this? I can't think of any sensible solution.

You could possibly treat an empty array on the Java side as null in the JSON serialisation and undefined as null in Java. That should allow to handle the three states.

spoenemann:
Yes, or use `Optional<List<WorkspaceFolder>>`. But for now I think we can handle both cases as null in Java because a language server probably doesn't really care about the distinction of _the client does not support workspace folders_ and _the client supports workspace folders but none are configured_.

PR #177, opened 2018-04-11
Implemented builder class for Launcher

spoenemann:
This patch has been extracted from #175 since it's unrelated to the original problem of supporting Hover.

The goal is to customize the JSON-RPC wiring more easily. I applied this customization approach to the DebugLauncher used by the debug protocol.

@jonahkichwacoders could you check this change regarding the debug protocol?


jonahkichwacoders:
I am away from my desk this week
I can review early next week.

Jonah

jonahkichwacoders:
@spoenemann I am looking at this now.

Issue #178, opened 2018-04-14
Implement textDocument/implementation


gorkem:
Implement the textDocument/implemention feature introduced with LSP 3.6
https://microsoft.github.io/language-server-protocol/specification#textDocument_implementation

Issue #118, opened 2017-09-07
Support for proposed color provider.


angelozerr:
VSCode  provides color provider for CSS and JSON that I have started to implement in bluesky with my PR https://github.com/mickaelistria/eclipse-bluesky/pull/43

VSCode has defined in their proposed https://github.com/Microsoft/vscode/blob/master/src/vs/vscode.proposed.d.ts this color provider feature:

```

/**
	 * The document color provider defines the contract between extensions and feature of
	 * picking and modifying colors in the editor.
	 */
	export interface DocumentColorProvider {
		/**
		 * Provide colors for the given document.
		 *
		 * @param document The document in which the command was invoked.
		 * @param token A cancellation token.
		 * @return An array of [color ranges](#ColorRange) or a thenable that resolves to such. The lack of a result
		 * can be signaled by returning `undefined`, `null`, or an empty array.
		 */
		provideDocumentColors(document: TextDocument, token: CancellationToken): ProviderResult<ColorRange[]>;
	}

```

It should be cool if lsp4j could support this feature too. For the moment I use lsp4j Range and my own structure and I cannot test colorProvider capabilities (see https://github.com/mickaelistria/eclipse-bluesky/pull/43/files#diff-e43b97ca17c7165ca88ef3cc690c9fecR61).

akosyakov:
There is now a node package containing the LSP definitions for typescript: https://github.com/Microsoft/vscode-languageserver-node/tree/9d6b53e0feb4039af95e2d108c376d8fc92d8b37/protocol

It contains the LSP as well as proposed extensions to it to support in VS code. We could add java interfaces for proposed extensions as well. E.g. for coloring: https://github.com/Microsoft/vscode-languageserver-node/blob/9d6b53e0feb4039af95e2d108c376d8fc92d8b37/protocol/src/protocol.colorProvider.proposed.ts

akosyakov:
PRs are welcomed.

An interface extending `LanguageServer` providing `textDocument/documentColor` request according to https://github.com/Microsoft/vscode-languageserver-node/blob/9d6b53e0feb4039af95e2d108c376d8fc92d8b37/protocol/src/protocol.colorProvider.proposed.ts should be put under https://github.com/eclipse/lsp4j/tree/master/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/proposed/colorProvider

@angelozerr Could you look into it?

angelozerr:
[...]

PR #179, opened 2018-04-16
Implemented changes to LSP versions 3.6 and 3.7

### LINKED ISSUE #167: Adapt to Protocol Level 3.6
https://github.com/Microsoft/language-server-protocol/blob/master/protocol-changelog.md

spoenemann:
Fixes #167, #178, and #118.

Unfortunately, this change massively breaks API. I don't know how else we could realize the change in LSP from

* method: ‘textDocument/completion’
* params: `TextDocumentPositionParams`

to

* method: ‘textDocument/completion’
* params: `CompletionParams`

In TypeScript this change is API compatible, in Java it's not :-(


org.eclipse.lsp4j/src/main/xtend-gen/org/eclipse/lsp4j/WorkspaceFolder.java
```diff
@@ -26,6 +30,18 @@
    */
   private String name;
```
akosyakov: In the spec it is not marked as optional. Is it bug there?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -1618,7 +1869,7 @@ class MarkupContent {
 	 * The type of the Markup.
 	 */
 	@NonNull
-	String MarkupKind;
+	String kind;
```
akosyakov: It is named as `kind` in the spec

jvalkeal:
Just out of curiosity if you could start over, what comes for a `textDocument/completion` and other objects, would you simply go with `TextDocumentCompletion` and then postfixing request/response classes with `Request` or `Response`? Then all changes would be internal to those classes.

org.eclipse.lsp4j/src/main/xtend-gen/org/eclipse/lsp4j/WorkspaceFolder.java
spoenemann: The LSP documentation says _"The name of the workspace folder. Defaults to the uri's basename."_ I think it's a bug in the specification: either the property should be optional, or the second part of the property documentation should be removed.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
spoenemann: That's why I corrected it here :-)

spoenemann:
@akosyakov I'll prepare a change log for the release.

@jvalkeal I don't understand what you mean. Could you explain further?

jonahkichwacoders:
While you are updating LS versions, it may be worth updating the Readme, it currently says:

### Supported LSP Versions

 * LSP4J 0.2.x &rarr; LSP 3.x
 * LSP4J 0.1.x &rarr; LSP 2.x

spoenemann:
Up to now we have only implemented the data types from the specification, and have not added any additional types.

Even then we wouldn't get around breaking the API. The root of the problem is that the LSP is designed for TypeScript, which has a completely different type system than Java. Some kinds of API changes that are 100% compatible in TypeScript break the corresponding Java representations.

spoenemann:
https://github.com/eclipse/lsp4j/blob/master/CHANGELOG.md

PR #211, opened 2018-07-24
Websockets support in new project to minimize disruption of other projects

juliandolby:


jonahgraham:
Thanks for the PR. A few things will need to be done before this can be merged into master.

- [ ] Remove writing to System.out/err, probably replace them with logging
- [ ] Add some tests
- [ ] Make this bidirectional, LSP has requests and notifications going in both directions.
- [ ] Split the JSON-RPC from the LSP specific parts of the code -- this is probably not required, but as my personal interest is getting the debug protocol to work over websockets, I need to have a websocket implementation that is not dependent on [LanguageServer](https://github.com/eclipse/lsp4j/blob/378e4e49b3e78e75623fab313b27c5c17c846388/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/LanguageServer.java) -- this is equivalent to how [Launcher](https://github.com/eclipse/lsp4j/blob/master/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/Launcher.java) is separate from [LSPLauncher](https://github.com/eclipse/lsp4j/blob/378e4e49b3e78e75623fab313b27c5c17c846388/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/launch/LSPLauncher.java)

Thank you for this starting point. Please let me know if you will be able to work on any of the above.

juliandolby:
Thanks for your comments.  As to your suggestions, I agree with them:

I definitely will do the first and will look into how to do tests with Tomcat.  

For being bidirectional, I do have messages going both ways with my connections to JupyterLab now.  The code uses LanguageClientAware to connect the other way.  Perhaps you could explain in more detail what you mean by bidirectional?

I agree with your fourth point, but would likely need some help to make it happen.

jonahgraham:
The bidirectional was about each message type being able to go both ways. AFAICT you are receiving request and notifcations from the remote, but is there a way to send a request to the remote and receive a response?

jonahgraham:
BTW the ip-validation is failing because your commits are not [Signed-off-by](https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#Signing_off_on_a_commit), plus if you have not already, you need to complete the [ECA](https://www.eclipse.org/legal/ECA.php). We can't merge that in until it is resolved, and the easiest way may be just to squash all your commits later and then sign it off.

org.eclipse.lsp4j.websockets/src/main/java/org/eclipse/lsp4j/launch/websockets/LSPWebSocketServer.java
```diff
@@ -0,0 +1,133 @@
...
+     
+        } else {
+        	assert false : "message " + m + " not understood";
+        	return null;
+        }
+    }
+
+    @OnError
+    public void onError(Throwable e, Session session){
+        e.printStackTrace();
```
jonahgraham: Thanks for doing the logging other places. Can you convert this one too.

jonahgraham:
> I agree with your fourth point, but would likely need some help to make it happen.

Lets keep this in mind as we go, but focus on the first three for now. It may be best to split this off into another issue.

juliandolby:
Thanks for pointing out all those issues.  I now just did the ECA, and my latest commit used -s to sign off.  In that commit, I logged the exception on error, which you are right I had missed before.

org.eclipse.lsp4j.websockets/src/main/java/org/eclipse/lsp4j/launch/websockets/LSPWebSocketServer.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @OnClose
+    public void onClose(Session session) throws IOException{
+        System.out.println("Close Connection " + session + " ...");
+        endpoints.remove(session);
+        session.close();
+    }
+
+    @OnMessage
+    public String onMessage(String message, Session session) throws InterruptedException, ExecutionException{
```
spoenemann: This method lacks error handling for messages (see StreamMessageProducer) and validation (ReflectiveMessageValidator).

org.eclipse.lsp4j.websockets/src/main/java/org/eclipse/lsp4j/launch/websockets/LSPWebSocketServer.java
```diff
@@ -0,0 +1,133 @@
...
+        private final Map<Session,Endpoint> endpoints = new LinkedHashMap<>();
+
+	private final MessageJsonHandler jsonHandler;
+	private final Supplier<T> newServer;
+	
+	LSPWebSocketServer(Supplier<T> newServer, Class<T> serverClass) {
+		this.newServer = newServer;
+		
+		Map<String, JsonRpcMethod> supportedMethods = new LinkedHashMap<>();
+		supportedMethods.putAll(ServiceEndpoints.getSupportedMethods(LanguageClient.class));
```
spoenemann: The remote service interface should be configurable.

org.eclipse.lsp4j.websockets/src/main/java/org/eclipse/lsp4j/launch/websockets/LSPWebSocketServer.java
```diff
@@ -0,0 +1,133 @@
...
+		supportedMethods.putAll(ServiceEndpoints.getSupportedMethods(serverClass));
+		
+		this.jsonHandler = new MessageJsonHandler(supportedMethods);
```
spoenemann: The JSON handler should be configurable.

org.eclipse.lsp4j.websockets/src/main/java/org/eclipse/lsp4j/launch/websockets/LSPWebSocketServer.java
```diff
@@ -0,0 +1,133 @@
...
+		Endpoint local = new GenericEndpoint(Collections.singleton(server));
+		endpoints.put(session, local);
+		
+		if (server instanceof LanguageClientAware) {
+			MessageConsumer out = new MessageConsumer() {
+				@Override
+				public void consume(Message arg0) throws MessageIssueException, JsonRpcException {
+				    LOG.log(Level.INFO, "sending message: " + arg0);
+					try {
+						session.getBasicRemote().sendText(arg0.toString());
```
spoenemann: Use the MessageJsonHandler to serialize the message.

org.eclipse.lsp4j.websockets/src/main/java/org/eclipse/lsp4j/launch/websockets/LSPWebSocketServer.java
```diff
@@ -0,0 +1,133 @@
...
+					} catch (IOException e) {
+						throw new Error(e);
```
spoenemann: Don't wrap checked exceptions using an Error. Use a RuntimeException instead.

org.eclipse.lsp4j.websockets/src/main/java/org/eclipse/lsp4j/launch/websockets/LSPWebSocketServer.java
```diff
@@ -0,0 +1,133 @@
...
+			};
+
+			RemoteEndpoint remote = new RemoteEndpoint(out, local);
+			((LanguageClientAware)server).connect(ServiceEndpoints.toServiceObject(remote, LanguageClient.class));
+		}
+    }
+
+    @OnClose
+    public void onClose(Session session) throws IOException{
+        System.out.println("Close Connection " + session + " ...");
```
spoenemann: Use the logger instead of System.out.println

org.eclipse.lsp4j.websockets/src/main/java/org/eclipse/lsp4j/launch/websockets/LSPWebSocketServer.java
```diff
@@ -0,0 +1,133 @@
...
+    @OnMessage
+    public String onMessage(String message, Session session) throws InterruptedException, ExecutionException{
+        Message m = jsonHandler.parseMessage(message);
+        Endpoint localEndpoint = endpoints.get(session);
+        
+        if (m instanceof RequestMessage) {
+            LOG.log(Level.INFO, "Request from the client: " + m);
+        	RequestMessage requestMessage = (RequestMessage)m;
+        	CompletableFuture<?> result = localEndpoint.request(requestMessage.getMethod(), requestMessage.getParams());
+        	Object r = result.join();
```
spoenemann: You need to catch CompletionException and CancellationException and handle them appropriately.

jonahgraham:
Ok to test

Issue #214, opened 2018-07-25
Support for DocumentSymbol 


yaohaizh:
Related changelog: 
https://microsoft.github.io/language-server-protocol/specification#version_3_10_0

Related issues: 
https://github.com/redhat-developer/vscode-java/issues/586

apupier:
@yaohaizh do you mean adding support for hierarchical document symbol? I think basic Document Symbol is available in LSp4J https://github.com/eclipse/lsp4j/blob/83ae40a1d4b2b7076a8bb8ea28a953cf1fb1ca84/org.eclipse.lsp4j/src/main/xtend-gen/org/eclipse/lsp4j/DocumentSymbolCapabilities.java#L19

yaohaizh:
@apupier Yes, it is about hierarchical document symbol. The response for ‘textDocument/documentSymbol’ is augmented with new value: **DocumentSymbol[]** | SymbolInformation[] | null

yaohaizh:
Add related issue: https://github.com/redhat-developer/vscode-java/issues/560

spoenemann:
Now blocked by #226.

PR #217, opened 2018-08-01
Add DocumentSymbol support.

yaohaizh:
Signed-off-by: Yaohai Zheng <yaozheng@microsoft.com>

https://github.com/eclipse/lsp4j/issues/214

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/TextDocumentService.java
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/TextDocumentService.java
@@ -32,6 +32,7 @@ import org.eclipse.lsp4j.DocumentLink;
 import org.eclipse.lsp4j.DocumentLinkParams;
 import org.eclipse.lsp4j.DocumentOnTypeFormattingParams;
 import org.eclipse.lsp4j.DocumentRangeFormattingParams;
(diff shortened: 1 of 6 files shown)


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/TextDocumentService.java
```diff
@@ -150,7 +151,7 @@
 	 * Registration Options: TextDocumentRegistrationOptions
 	 */
 	@JsonRequest
-	CompletableFuture<List<? extends SymbolInformation>> documentSymbol(DocumentSymbolParams params);
+	CompletableFuture<Either<List<? extends SymbolInformation>, List<? extends DocumentSymbol>>> documentSymbol(DocumentSymbolParams params);
```
spoenemann: I doubt that the current implementation of Either is able to parse this return type. Could you add a test to JsonParseTest that checks parsing a `documentSymbol` ResponseMessage? If it fails, please mark it with `@Ignore` and create an issue for that problem. Then we can merge this PR anyway and take care of the parsing later.

yaohaizh:
@spoenemann I have add testcase for DocumentSymbol. But not sure if it is the right format/content for test case since it is failed anyway.

PR #223, opened 2018-08-09
GH-221: Added the encode/decode utility for the highlighting tokens.

### LINKED ISSUE #221: Provide a utility class to encode/decode the binary representation of a semantic highlighting token
The binary format was proposed here: 
https://github.com/Microsoft/vscode-languageserver-node/pull/367#issuecomment-403085124.

The implementation for the JDT LS is available here: https://github.com/eclipse/eclipse.jdt.ls/pull/746/files#diff-cdffa7f7d74681bc62d80a7dea2c5452.

I have started to work on [the support for Xtext](https://github.com/eclipse/xtext-core/issues/812) and now, I have to copy over the logic. Having a utility class in LSP4J would be useful.

CC: @spoenemann, @svenefftinge.

kittaakos:
Closes #221.

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/util/SemanticHighlightingTokens.java
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/util/SemanticHighlightingTokens.java
@@ -1,0 +1,237 @@
+/*******************************************************************************
+ * Copyright (c) 2018 TypeFox GmbH (http://www.typefox.io) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.lsp4j.util;
+
+import static java.util.Collections.emptyList;
+
+import java.nio.ByteBuffer;
+import java.util.Base64;
+import java.util.List;
+
+import org.eclipse.xtext.xbase.lib.IterableExtensions;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ComparisonChain;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableList.Builder;
+import com.google.common.collect.Iterables;
+
+/**
+ * Utility class for encoding and decoding semantic highlighting tokens into a
+ * compact, {@code base64} representation.
+ */
+public final class SemanticHighlightingTokens {
+
+	private static int LENGTH_SHIFT = 0x0000010;
+	private static int SCOPES_MASK = 0x0000FFFF;
+
+	/**
+	 * Encodes the iterable of tokens into a compact, {@code base64} string. Returns
+	 * with an empty string if the {@code tokens} argument is {@code null} or empty.
+	 */
+	public static String encode(Iterable<? extends Token> tokens) {
+		if (IterableExtensions.isNullOrEmpty(tokens)) {
+			return "";
+		}
+		// 2 stands for: number of elements for the output; the "character" and the
+		// "length and scope".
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/util/SemanticHighlightingTokens.xtend
```diff
@@ -0,0 +1,172 @@
...
+
+		/**
+		 * Converts an iterable of tokens into an array of primitive integers. Returns with
+		 * an empty array if the argument is {@code null} or empty.
+		 * 
+		 * <p>
+		 * <ul>
+		 *   <li>{@code array[3*i]} is the {@link Token#character character} of the token.</li>
+		 *   <li>{@code array[3*i+1]} is the {@link Token#length length} of the token.</li>
+		 *   <li>{@code array[3*i]} is the token {@link Token#scope scope}.</li>
```
kittaakos: Incorrect doc. Should be `3*i+2`.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/util/SemanticHighlightingTokens.xtend
```diff
@@ -0,0 +1,172 @@
...
+			}
+			return array;
+		}
+
+		/**
+		 * Counter-part of the {@link Token#toIntArray}. Converts an array of primitive integers
+		 * into a list of tokens. If the input is {@code null}, returns with an empty list.
+		 * Throws an exception if the length of the arguments is not a modulo of {@code 3}.
+		 */
+		static def fromIntArray(int... input) {
```
kittaakos: I need to an assertion here. The ints cannot exceed the 0 and 2<sup>32</sup>-1 (inclusive) range.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/util/SemanticHighlightingTokens.xtend
```diff
@@ -0,0 +1,172 @@
...
+
+	static val LENGTH_SHIFT = 0x0000010;
+	static val SCOPES_MASK = 0x0000FFFF;
+
+	/**
+	 * Encodes the iterable of tokens into a compact, {@code base64} string.
+	 * Returns with an empty string if the {@code tokens} argument is {@code null} or
+	 * empty.
+	 */
+	static def String encode(Iterable<? extends Token> tokens) {
```
kittaakos: I have to add a encode/decode test with a huge chunk of data.

kittaakos: [Done](https://github.com/eclipse/lsp4j/pull/223/files#diff-d1a1aa035a627187770faebecd6783e7R145).

kittaakos:
Ready for review.

svenefftinge:
We use Xtend only for protocol definitions. In other places please use Java.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/util/SemanticHighlightingTokens.java
```diff
@@ -0,0 +1,237 @@
...
+import com.google.common.collect.ComparisonChain;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableList.Builder;
+import com.google.common.collect.Iterables;
+
+/**
+ * Utility class for encoding and decoding semantic highlighting tokens into a
+ * compact, {@code base64} representation.
+ */
+public abstract class SemanticHighlightingTokens {
```
spoenemann: Usually I make such a class `final`  and declare a private default constructor instead of making it abstract. An abstract class looks like something that needs to be implemented in a subclass.

PR #231, opened 2018-08-17
Aligned the protocol implementation with version 3.10.0

### LINKED ISSUE #169: Support for proposed folding provider.
VSCode provides folding provider for HTML, JSON.

VSCode has defined in their proposed https://github.com/Microsoft/vscode/blob/master/src/vs/vscode.proposed.d.ts#L37

See https://github.com/Microsoft/language-server-protocol/issues/418

It should be cool if lsp4j could provide this feature. Thanks!

spoenemann:
Fixes #169.

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -452,6 +452,11 @@ class DocumentSymbolCapabilities extends DynamicRegistrationCapabilities {
      */
     SymbolKindCapabilities symbolKind
     
+    /**
+     * The client support hierarchical document symbols.
+     */
+    Boolean hierarchicalDocumentSymbolSupport
+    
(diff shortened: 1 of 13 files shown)


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -897,13 +929,10 @@ class CodeAction {
     new() {
     }
     
-    new(@NonNull String title, String kind, List<Diagnostic> diagnostics, WorkspaceEdit edit, Command command) {
+    new(@NonNull String title) {
```
akosyakov: Won't it be a breaking change? Or we don't care?

spoenemann: The class is new in 0.5.0. I aligned it with the rest of our implementation, where we specify constructors only when there are only few properties.

PR #235, opened 2018-08-18
DAP: Events use 'event' key instead of 'command'

### LINKED ISSUE #234: Debug Adapter Protocol: Events do not have 'event' key
A sample JSON event output from my debug adapter:

```json
{"type":"event","seq":3,"command":"initialized","body":null}
```

The specification - however - requires an `event` key, instead of a `command` key:

```typescript
interface Event extends ProtocolMessage {
  type: 'event';

  /**
   * Type of event.
   */
  event: string;

  /**
   * Event-specific information.
   */
  body?: any;
}

interface InitializedEvent extends Event {
  event: 'initialized';
}
```
See https://microsoft.github.io/debug-adapter-protocol/specification#Events_Initialized

Not sure if this is a version inconsistency related to #232

fwcd:
Fixes #234.


fwcd:
I will probably have to sign my commits prior to merging.

jonahgraham:
This test now fails, it has an (incorrectly) encoded event in it still org.eclipse.lsp4j.jsonrpc.debug.test.DebugLauncherTest.testCustomGson()

fwcd:
@jonahgraham This is strange, DebugLauncherTest is being skipped when running all tests on my machine...

I will investigate this one further.

PR #238, opened 2018-08-22
Omit "arguments" if empty on DebugMessageTypeAdapter

drungrin:
Some endpoints throws a NPE on methods without parameters. Per JSON-RPC 2.0 Specification, If present, parameters for the rpc call MUST be provided as a Structured value.


spoenemann:
Could you [create an issue](https://github.com/eclipse/lsp4j/issues) and explain a bit more the scenario where the current implementation leads to an incompatibility with the JSON-RPC specification?

Your commit did not pass IP validation because it lacks the signed-off-by signature. You can add it with `git commit --amend -s`.

drungrin:
As requested, fixed the signed-off-by, the user.name should be my github username instead of my name :)

drungrin:
Updated as proposed.

jonahgraham:
@drungrin Thanks for the fix.

Issue #236, opened 2018-08-18
Add builds / reporting for PRs


jonahgraham:
I [have requested](https://bugs.eclipse.org/bugs/show_bug.cgi?id=538053) GitHub PR plug-in for Jenkins so that our PRs can be tested before merging (It would have made it easier for @fwcd on #235 as it would have reported the failing test automatically).

I recently did this for Nebula, so, once added to Jenkins I should be able to get it up and running quickly.

jonahgraham:
Done: https://github.com/eclipse/lsp4j/blob/master/Contributing.md#jenkins-pr-build

PR #242, opened 2018-08-23
Issue #236: Remove Jenkinsfile for PR builds

jonahgraham:
I am not using this in the end.


jonahgraham:
If someone with more knowledge can get the GitHub PR Jenkins plug-in + Pipeline working, have at it. For now the configuration is pretty simple (just run gradle) that the cost of not using Jenkinsfile is lowish.

Issue #240, opened 2018-08-23
Review using null for fields with no values


jonahgraham:
Follow up to #239. @drungrin in #238 fixed the `"argument"` part of [this](https://github.com/eclipse/lsp4j/issues/239#issuecomment-415422626), the rest needs a closer review:

> I think the correct place to make this change is in DebugMessageTypeAdapter's write method to omit the "arguments" if the getParams returns null. I suspect the same thing may apply to "body" of notifications and responses and "message" of error responses, all of those within the same write method.

drungrin:
The body is omitted on responses in Visual Studio Code. On notifications threre is always a body, since the only notification without argument is "initialized", which is sent by the Java Debug Server with a body. I believe it is safe to omit the body in the notifications also.

`"{"event":"initialized","body":{"type":"initialized"},"seq":2,"type":"event"}"`

`"{"success":true,"request_seq":2,"command":"launch","seq":3,"type":"response"}"`

jonahgraham:
I don't think this issue needs to hold up the 0.5.0 release, updating milestone to 0.6.0.

jonahgraham:
@drungrin Thanks for following up.

PR #246, opened 2018-08-23
Review using null for fields with no values #240 

drungrin:
The body is omitted on responses in Visual Studio Code. On notifications threre is always a body, since the only notification without argument is "initialized", which is sent by the Java Debug Server with a body. I believe it is safe to omit the body in the notifications also.

### DIFF
--- a/org.eclipse.lsp4j.jsonrpc.debug/src/main/java/org/eclipse/lsp4j/jsonrpc/debug/adapters/DebugMessageTypeAdapter.java
+++ b/org.eclipse.lsp4j.jsonrpc.debug/src/main/java/org/eclipse/lsp4j/jsonrpc/debug/adapters/DebugMessageTypeAdapter.java
@@ -448,12 +448,11 @@ public class DebugMessageTypeAdapter extends MessageTypeAdapter {
 			} else {
 				out.name("success");
 				out.value(true);
-				out.name("body");
 				Object result = responseMessage.getResult();
-				if (result == null)
-					writeNullValue(out);
-				else
+				if (result != null) {
+					out.name("body");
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4j.jsonrpc.debug/src/main/java/org/eclipse/lsp4j/jsonrpc/debug/adapters/DebugMessageTypeAdapter.java
```diff
@@ -463,12 +462,11 @@ public void write(JsonWriter out, Message message) throws IOException {
 			writeIntId(out, notificationMessage.getRawId());
 			out.name("event");
 			out.value(notificationMessage.getMethod());
-			out.name("body");
-			Object params = notificationMessage.getParams();
-			if (params == null)
-				writeNullValue(out);
-			else
+      Object params = notificationMessage.getParams();
```
jonahgraham: Bad leading whitespace on this line.

jonahgraham:
ok to test

jonahgraham:
@drungrin thanks for picking this up and running with it.

jonahgraham:
PS The new CI check (#236) failed: "Build finished. 244 tests run, 0 skipped, 1 failed."

drungrin:
I don't think this test represents the real scenario, since the payloads from vscode ommits the body as exposed on #240

What do you guys think?

jonahgraham:
Absolutely correct! Please update the test to represent reality in this case.

jonahgraham:
I am on holiday and won't be able to merge this in until midday Monday at the earliest.

drungrin:
Updated as requested.

Issue #259, opened 2018-09-24
Server interface implementation has no way to discover what client is making a request


robstryker:
When a new request comes in from a client, a server interface has no method or API to discover which client is making the request.  The stack trace looks as follows:

ThreadClientIntegrationTest$MyServerImpl.askServer(MyParam) line: 79	
NativeMethodAccessorImpl.invoke0(Method, Object, Object[]) line: not available [native method]	
NativeMethodAccessorImpl.invoke(Object, Object[]) line: 62	
DelegatingMethodAccessorImpl.invoke(Object, Object[]) line: 43	
Method.invoke(Object, Object...) line: 564	
GenericEndpoint.lambda$1(AnnotationUtil$MethodInfo, Object, Object) line: 65	
1084502906.apply(Object) line: not available	
GenericEndpoint.request(String, Object) line: 120	
RemoteEndpoint.handleRequest(RequestMessage) line: 261	
RemoteEndpoint.consume(Message) line: 190	
StreamMessageProducer.handleMessage(InputStream, StreamMessageProducer$Headers) line: 192	
StreamMessageProducer.listen(MessageConsumer) line: 94	
ConcurrentMessageProcessor<T>.run() line: 122	
Executors$RunnableAdapter<T>.call() line: 514	
FutureTask<V>.run() line: 264	

We can see that the ConcurrentMessageProcessor is essentially driving the communication between the message producer (the client) and the consumer (the server's implementation class). But this processor has no ability to directly have a reference to the client remote proxy that matches the given producer making the request. 

As an example, if we take the https://github.com/TypeFox/lsp4j-chat-app  chat-app from typefox...  and we imagine one client continues to spam the channel, the server may want to take steps to censure him, rate limit his messaging, mute him, or outright close the connection. 

However, the ChatServerImpl class  (https://github.com/TypeFox/lsp4j-chat-app/blob/master/src/main/java/io/typefox/lsp4j/chat/server/ChatServerImpl.java)   has no way to discover which client is making the request. 

As another example, if we assume a user must authenticate themselves to get a given name, the typefox example is forced to have the client send their name (or if authentication is required, name and password) with each message. If the server implementation could accurately map the given stream to a client, it wouldn't need to do this. It could prompt for the username and password only once, and simply keep the data until the stream is closed. 

[...]

PR #260, opened 2018-09-24
[issue 259] Allow server implementations to discover what client make a given request

robstryker:
Signed-off-by: Rob Stryker <stryker@redhat.com>

### DIFF
--- a/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/Launcher.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/Launcher.java
@@ -293,6 +293,7 @@ public interface Launcher<T> {
 
 		@SuppressWarnings("unchecked")
 		public Launcher<T> create() {
+			// Validate input
 			if (input == null)
 				throw new IllegalStateException("Input stream must be configured.");
 			if (output == null)
@@ -302,6 +303,7 @@ public interface Launcher<T> {
 			if (remoteInterfaces == null)
 				throw new IllegalStateException("Remote interface must be configured.");
 			
+			// Create proxy and endpoints
 			MessageJsonHandler jsonHandler = createJsonHandler();
 			RemoteEndpoint remoteEndpoint = createRemoteEndpoint(jsonHandler);
 			T remoteProxy;
@@ -310,28 +312,61 @@ public interface Launcher<T> {
 			} else {
 				remoteProxy = (T) ServiceEndpoints.toServiceObject(remoteEndpoint, (Collection<Class<?>>) (Object) remoteInterfaces, classLoader);
 			}
+			
+			// create the message processor
 			StreamMessageProducer reader = new StreamMessageProducer(input, jsonHandler, remoteEndpoint);
 			MessageConsumer messageConsumer = wrapMessageConsumer(remoteEndpoint);
 			ExecutorService execService = executorService != null ? executorService : Executors.newCachedThreadPool();
+			ConcurrentMessageProcessor<T> msgProcessor = createMessageProcessor(reader, messageConsumer, remoteProxy);
 			
-			return new Launcher<T> () {
-				@Override
-				public Future<Void> startListening() {
-					return ConcurrentMessageProcessor.startProcessing(reader, messageConsumer, execService);
-				}
-
-				@Override
-				public T getRemoteProxy() {
-					return remoteProxy;
-				}
-
-				@Override
-				public RemoteEndpoint getRemoteEndpoint() {
-					return remoteEndpoint;
-				}
-			};
+			return createLauncher(reader, messageConsumer, execService, remoteProxy, remoteEndpoint, msgProcessor);
 		}
 		
+		protected Launcher<T> createLauncher(StreamMessageProducer reader, MessageConsumer messageConsumer,
+				ExecutorService execService, T remoteProxy, RemoteEndpoint remoteEndpoint, ConcurrentMessageProcessor<T> msgProcessor) {
+			return new StandardLauncher<T>(reader, execService, remoteProxy, remoteEndpoint, msgProcessor);
+		}
+		
+		public static class StandardLauncher<T> implements Launcher<T> {
+			private StreamMessageProducer reader;
+			private ExecutorService execService;
+			private T remoteProxy;
+			private RemoteEndpoint remoteEndpoint;
+			private ConcurrentMessageProcessor<T> msgProcessor;
+
+			public StandardLauncher(StreamMessageProducer reader, MessageConsumer messageConsumer,
+					ExecutorService execService, T remoteProxy, RemoteEndpoint remoteEndpoint) {
+				this(reader, execService, remoteProxy, remoteEndpoint, 
+						new ConcurrentMessageProcessor<>(reader, messageConsumer));
+			}
+			
+			public StandardLauncher(StreamMessageProducer reader2,
+					ExecutorService execService2, T remoteProxy2, RemoteEndpoint remoteEndpoint2,
+					ConcurrentMessageProcessor<T> msgProcessor) {
+				this.reader = reader2;
+				this.execService = execService2;
+				this.remoteProxy = remoteProxy2;
+				this.remoteEndpoint = remoteEndpoint2;
+				this.msgProcessor = msgProcessor;
+			}
+
+			@Override
+			public Future<Void> startListening() {
+				final Future<?> result = execService.submit(msgProcessor);
+				return ConcurrentMessageProcessor.wrapFuture(result, reader);
+			}
+
+			@Override
+			public T getRemoteProxy() {
+				return remoteProxy;
+			}
+
+			@Override
+			public RemoteEndpoint getRemoteEndpoint() {
+				return remoteEndpoint;
+			}
+		}
+
 		protected MessageConsumer wrapMessageConsumer(MessageConsumer consumer) {
 			MessageConsumer result = consumer;
 			if (messageTracer != null) {
@@ -383,6 +418,22 @@ public interface Launcher<T> {
 				return new MessageJsonHandler(supportedMethods, configureGson);
 			else
 				return new MessageJsonHandler(supportedMethods);
+		}
+		
+		/**
+		 * Create a message processor given the provided parameters. 
+		 * 
+		 * Clients may override this method to create a message processor
+		 * with an expanded feature set.
+		 * 
+		 * @param reader
+		 * @param messageConsumer
+		 * @param remoteProxy
+		 * @return
+		 */
+		protected ConcurrentMessageProcessor<T> createMessageProcessor(MessageProducer reader, 
+				MessageConsumer messageConsumer, T remoteProxy) {
+			return  new ConcurrentMessageProcessor<T>(reader, messageConsumer);
 		}
 		
 		/**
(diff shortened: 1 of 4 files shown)


robstryker:
https://github.com/eclipse/lsp4j/issues/259

jonahgraham:
ok to test

mickaelistria:
Why would there be a 1 client/thread mapping? If you think about a possible LS using Vert.x message dispatcher, Vert.x doesn't create many threads and still is highly concurrent.

robstryker:
It's not really a 1:1 mapping so much as that a Thread can only be handling a request for 1 client at a time. It could be the case that two threads are handling the same client around the same time, and that should be fine, as this model asks what is the current client for the current thread. Remember, a server extender / implementer will receive a request (server/sendMessage) and simply want to know which client (remoteProxy) is making that request.   

I think this is a fair methodology. Threads can have high turnover or change who their client is, so long as they remember to update the threadlocal variables before calling the server's interface, and so long as that threadlocal variable is correct at the time the server implementation is called, and that the call to the server implementation request is synchronous and in the same thread. 

Even if it's not, and even if a message dispatcher somehow calls the server's method in a brand new thread, that thread could first update it's threadlocal value to know who the requester is. 

The real requirement here is just that some API exists for a server implementation to find out which remoteProxy is making a given request.  

@mickaelistria  asked the following:

> For request/response, we can assume that only the client which throws the request is interested in the response. What about notifications? They're sent to all clients?

This enhancement is not expected to change anything about the relationship between request/response vs notification. This enhancement only allows the server implementation to know who is making the request. 

As an example, let's imagine one user in a chat application wants "batch notifications" and another user wants live notifications. Perhaps user 1 will send a message like "/batchnotifications 5min".   The server implementation will want to know which client is requesting to switch from live to batch notifications. Later, when other users send messages, the server will be able to check it's models for which clients have requested batched notifications.   

In the current system, we cannot even associate the "/batchnotifications 5min" message with any specific user or client, so we cannot even update our models to customize behavior for them.

I think being able to know which client is making a request is a pretty normal requirement, and I think the implementation I have provided here stored in the Launcher is the best place for it for others to implement as well.

robstryker:
I've updated the PR to undo the generated file edit. Not sure how that snuck into the patch. The PR is much smaller now and more clear / concise.

mickaelistria:
Ok, I understand better the use case, and the issue, and the proposal.
So it's more a N-clients 1-server mapping you are interested in, and with the server being aware of which client is sending the request and wants a response, and which clients to send the notifications.

I'm thinking we're reaching the limits of the specification here.
Are you aware of existing clients/servers who already have tackled this multiplicity problem and can be studied for consideration? At some point, maybe the simplest would be to make the spec evolve to allow a client id in the messages or headers.
I also have the impression that threads are not the best way to categorize clients here. Can we assume that 1 client == 1 stream instead? How do you plan to connect the LS with the various clients anyway; all send content to the same stream but as different threads? Is there any guarantee that the message won't mess up on the stream?
Let's imagine the server is a node.js based server, how would you proposal about threads translate there? How would the node.js based LS understand that messages come from different clients?
...

robstryker:
> How would the node.js based LS understand that messages come from different clients?

So my main proposal is pretty simple. First, none of what I'm doing (I believe) affects the spec at all. This is purely an implementation-detail on the lsp4j.jsonrpc implementation. 

At some point, some code is reading a message from a stream and calling the server interface's implementation. This, necessarily, must be done inside one thread or another, right?  Eventually they end up calling a LS's method corresponding to the entry point specified in the message.  Since lsp4j is calling the interface via some reflection / lambda / etc, there's no ability to add parameters to the server impl's interface (nor would it be desirable, in my opinion). 

So, RIGHT before calling the actual interface method on the server, *something* should be able to set in the model which client (aka remote proxy) aligns with this request... which client is making this request. 

My PR puts this in ConcurrentMessageProcessor because it was the easiest class in my opinion.  However I also could have put it somewhere in RemoteEndpoint.consume(etc), which would be the closest method to the server implementation class's methods being called.  Let's look at the stacktrace here:

```
ThreadClientIntegrationTest$MyServerImpl.askServer(MyParam) line: 79	
NativeMethodAccessorImpl.invoke0(Method, Object, Object[]) line: not available [native method]	
NativeMethodAccessorImpl.invoke(Object, Object[]) line: 62	
DelegatingMethodAccessorImpl.invoke(Object, Object[]) line: 43	
Method.invoke(Object, Object...) line: 564	
GenericEndpoint.lambda$1(AnnotationUtil$MethodInfo, Object, Object) line: 65	
1083021083.apply(Object) line: not available	
GenericEndpoint.request(String, Object) line: 120	
RemoteEndpoint.handleRequest(RequestMessage) line: 261	
RemoteEndpoint.consume(Message) line: 190	
StreamMessageProducer.handleMessage(InputStream, StreamMessageProducer$Headers) line: 192	
StreamMessageProducer.listen(MessageConsumer) line: 94	
ConcurrentMessageProcessor<T>.run() line: 122	
```

My current implementation stores a ThreadLocal variable near the start of this process, near ConcurrentMessageProcessor:122.   This implementation works because the lsp4j.jsonrpc launcher is reading / writing to one set of streams, for one launcher, in one thread. It's a dedicated thread reading messages from one set of streams with one remoteProxy representing the client. However, if we REALLY wanted, we could put it around RemoteEndpoint.consume(Message) line: 190, right before we reflection our way into the server implementation. 

The bulk of this patch is really two small things. I move the Builder out of the Launcher class, just for clarity. This could probably be undone. I also create some inner classes so that we're not returning on-the-fly anonymous inner classes, again for clarity. And I make ConcurrentMessageProcessor a generic, so it can receive the remoteProxy object. 

When we run() the ConcurrentMessageProcessor,  I just make sure to store in a ThreadClientMap that this thread (which is reading one socket and funneling the message to be handled) is associated with which remote proxy object. When a server implementation finally gets a call to do something, it now has some model where it can get the current client making the request. 

If i undid some of the separate-out-for-clarity stuff, this patch would be way smaller.

robstryker:
Thanks for the great feedback. I was really just trying to get any feedback / discussion on this issue. 

I realize lsp4j / language server protocols has a one-client-one-server "assumption", and that each LSP will inherit that assumption. However, lsp4j.jsonrpc may one day be used in more varied protocols, so any discussion that makes this extensible makes me happy. 

I'll try to update this PR today.

robstryker:
PR updated and ready for review. Let me know if there's any API issues in terms of binary incompatibilities this may create. I don't believe this should cause any issues, but if it does I can fix them.

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/Launcher.java
```diff
@@ -385,6 +420,22 @@ protected MessageJsonHandler createJsonHandler() {
...
+		/**
+		 * Create a message processor given the provided parameters. 
+		 * 
+		 * Clients may override this method to create a message processor
+		 * with an expanded feature set.
+		 * 
+		 * @param reader
+		 * @param messageConsumer
+		 * @param remoteProxy
+		 * @return
```
spoenemann: Please remove such empty JavaDoc tags.

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/ConcurrentMessageProcessor.java
```diff
@@ -27,20 +27,27 @@
 /**
  * This class connects a message producer with a message consumer by listening for new messages in a dedicated thread.
  */
-public class ConcurrentMessageProcessor implements Runnable {
+public class ConcurrentMessageProcessor<T> implements Runnable {
 	
 	/**
 	 * Start a thread that listens for messages in the message producer and forwards them to the message consumer.
 	 * 
+	 * This method is deprecated. Please use the non-static ConcurrentMessageProcessor.beginProcessing() instead.
```
spoenemann: This explanation should be moved to the `@deprecated` tag

org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/ExtendableConcurrentMessageProcessorTest.java
```diff
@@ -0,0 +1,266 @@
+/******************************************************************************
+ * Copyright (c) 2016 TypeFox and others.
```
spoenemann: Copyright statement should be updated for this new file (year + initial contributor affiliation).

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/Launcher.java
```diff
@@ -310,28 +312,61 @@
...
 
-				@Override
-				public RemoteEndpoint getRemoteEndpoint() {
-					return remoteEndpoint;
-				}
-			};
+			@Override
+			public Future<Void> startListening() {
+				final Future<?> result = execService.submit(msgProcessor);
+				return ConcurrentMessageProcessor.wrapFuture(result, reader);
```
spoenemann: You could call the new `beginProcessing(…)` method here (otherwise we could as well remove it).

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/ConcurrentMessageProcessor.java
```diff
@@ -27,20 +27,27 @@
 /**
  * This class connects a message producer with a message consumer by listening for new messages in a dedicated thread.
  */
-public class ConcurrentMessageProcessor implements Runnable {
+public class ConcurrentMessageProcessor<T> implements Runnable {
```
spoenemann: The new type parameter `T` is not used anywhere, is it?

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/ConcurrentMessageProcessor.java
```diff
@@ -27,20 +27,27 @@
...
 	 * 
+	 * This method is deprecated. Please use the non-static ConcurrentMessageProcessor.beginProcessing() instead.
+	 * 
 	 * @param messageProducer - produces messages, e.g. by reading from an input channel
 	 * @param messageConsumer - processes messages and potentially forwards them to other consumers
 	 * @param executorService - the thread is started using this service
 	 * @return a future that is resolved when the started thread is terminated, e.g. by closing a stream
+	 * @deprecated
 	 */
 	public static Future<Void> startProcessing(MessageProducer messageProducer, MessageConsumer messageConsumer,
```
spoenemann: Please add the `@Deprecated` annotation.

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/Launcher.java
```diff
@@ -310,28 +312,60 @@
...
+				ExecutorService execService, T remoteProxy, RemoteEndpoint remoteEndpoint, ConcurrentMessageProcessor msgProcessor) {
+			return new StandardLauncher<T>(reader, execService, remoteProxy, remoteEndpoint, msgProcessor);
+		}
+		
+		public static class StandardLauncher<T> implements Launcher<T> {
+			private StreamMessageProducer reader;
+			private ExecutorService execService;
+			private T remoteProxy;
+			private RemoteEndpoint remoteEndpoint;
+			private ConcurrentMessageProcessor msgProcessor;
```
spoenemann: The fields can be marked `final`.

robstryker:
Hi:

Suggested changes have been made

Issue #261, opened 2018-09-26
[enhancement] Provide better utilities for LSP types


kittaakos:
- (`DocumentSymbol`) -> `iterator` using a breadth-first traversal. Maybe pre-order and post-order too.
-  (`Range`, `Range`) -> `containsRange`, `isOverlappedBy`, `isAfterRange`, `isBeforeRange`, ...
- (`Range`, `Position`) -> `containsPosition`, `isAfter`, `isBefore`, ...

KamasamaK:
Some important ones are already covered in the opening post, but for reference, here is what the VS Code API provides for [Position](https://github.com/Microsoft/vscode/blob/a5580c3d51d083c738f7a1df952ac51cc5c006a5/src/vs/vscode.d.ts#L251) and [Range](https://github.com/Microsoft/vscode/blob/a5580c3d51d083c738f7a1df952ac51cc5c006a5/src/vs/vscode.d.ts#L370). I have found most of those methods very useful.

kittaakos:
> here is what the VS Code API provides for [Position](https://github.com/Microsoft/vscode/blob/a5580c3d51d083c738f7a1df952ac51cc5c006a5/src/vs/vscode.d.ts#L251) and [Range](https://github.com/Microsoft/vscode/blob/a5580c3d51d083c738f7a1df952ac51cc5c006a5/src/vs/vscode.d.ts#L370).

I was looking for precisely this, did not find it. Thanks for sharing!

KamasamaK:
Actually, implementing `java.lang.Comparable` would probably obviate the need for some comparison methods like `isBefore` and `isAfter`.

PR #263, opened 2018-10-02
GH-261: Added utilities for `Range`, `Position`, and `DocumentSymbol`.

kittaakos:
Closes: #261.


jonahgraham:
add to whitelist

org.eclipse.lsp4j/src/test/java/org/eclipse/lsp4j/test/util/PositionsTest.java
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	@Test(expected = NullPointerException.class)
+	public void isBefore_nullRight() {
+		Positions.isBefore(new Position(0, 0), null);
+	}
+
+	@Test
+	public void isBefore_equals() {
+		Assert.assertFalse(Positions.isBefore(new Position(3, 3), new Position(3, 3)));
```
kittaakos: This is incorrect.

org.eclipse.lsp4j/src/test/java/org/eclipse/lsp4j/test/util/PositionsTest.java
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	@Test
+	public void isBefore_same() {
+
```
kittaakos: TODO: remove LF.

org.eclipse.lsp4j/src/test/java/org/eclipse/lsp4j/test/util/PositionsTest.java
svenefftinge: Why? It looks good to me?

kittaakos: It is checking "same" ranges, not "equals". Technically, they're the same though.

PR #268, opened 2018-10-08
Add CodeActionOptions support.

yaohaizh:
Signed-off-by: Yaohai Zheng <yaozheng@microsoft.com>


dhuebner:
@yaohaizh Could you please re-base on master to get the latest Either type handling changes?

dhuebner:
@yaohaizh Thanks!
Unfortunately it doesn't seems like the current active annotation processor implementation was used. Could you check that the code is compiled with the current master state? E.g. `public void setCodeActionProvider(final Either<Boolean, CodeActionOptions> codeActionProvider) {...}` should have a null check before setting the value

yaohaizh:
@dhuebner Thanks. Updated the PR with master changes, and now the null check is added.

PR #273, opened 2018-10-09
GH-272: Updated the API to support type hierarchies.

kittaakos:
Added a method
 - for retrieving type hierarchy items from the server, and another 
 - for resolving any unresolved type hierarchy items.

Closes: #272

Signed-off-by: Akos Kitta <kittaakos@typefox.io>

PR #297, opened 2019-01-24
GH-296: Extended the protocol with `textDocument/callHierarchy`.

### LINKED ISSUE #296: Support for call hierarchies
The proposed LSP extension is [here](https://github.com/Microsoft/vscode-languageserver-node/pull/420).

kittaakos:
Closes #296.

----

LSP PR: https://github.com/Microsoft/vscode-languageserver-node/pull/420

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/TypeHierarchyDirection.java
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/TypeHierarchyDirection.java
@@ -1,0 +1,57 @@
+/******************************************************************************
+ * Copyright (c) 2019 TypeFox and others.
+ * 
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0,
+ * or the Eclipse Distribution License v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ * 
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ ******************************************************************************/
+package org.eclipse.lsp4j;
+
+import org.eclipse.lsp4j.services.TextDocumentService;
+
+/**
+ * Direction specific to the
+ * {@link TextDocumentService#typeHierarchy(TypeHierarchyParams)
+ * textDocument/typeHierarchy} and
+ * {@link TextDocumentService#resolveTypeHierarchy(ResolveTypeHierarchyItemParams)
+ * typeHierarchy/resolve} LS methods.
+ * 
+ * <p>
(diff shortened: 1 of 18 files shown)


kittaakos:
> should we also add some remark in their JavaDoc comments?

I will add some additional notes to the Javadoc.

(Will do the same for the [`textDocument/typeHierarchy` related changes](https://github.com/eclipse/lsp4j/pull/273) too.)

kittaakos:
I have documented the caveats between LSP4J and LSP for the following language features:
 - `textDocument/semanticHighlighting`,
 - `textDocument/typeHierarchy`, and
 - `textDocument/callHierarchy`.

spoenemann:
Great! :+1:

kittaakos:
I had to update the `typeHierarchy` API based on the LSP [feedback](https://github.com/Microsoft/vscode-languageserver-node/pull/426#pullrequestreview-196406599).

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/TypeHierarchyDirection.java
```diff
@@ -0,0 +1,57 @@
...
+ * <p>
+ * Valid values are:
+ * <ul>
+ * <li>{@link TypeHierarchyDirection#Children Children},</li>
+ * <li>{@link TypeHierarchyDirection#Parents Parents},</li>
+ * <li>{@link TypeHierarchyDirection#Both Both}.</li>
+ * </ul>
+ *
+ */
+public enum TypeHierarchyDirection {
```
spoenemann: The enum needs a field 'value' for the conversion from/to numbers to work, see other enums in the project.

The conversion is defined in [EnumTypeAdapter](https://github.com/eclipse/lsp4j/blob/master/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/EnumTypeAdapter.java).

kittaakos: Hold up a sec, what about [`TextDocumentSyncKind`](https://github.com/eclipse/lsp4j/blob/db9f9bdda89a67384f5c78985b4d3f89a001a92d/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/TextDocumentSyncKind.java#L17). My understanding was, we need the `value` field if the enum starts from a **non-zero** number bu we can use zero if the enum is optional on the use site. Can you please clarify? Thanks!

spoenemann: Right, the type adapter falls back to the ordinal value. Never mind, then.

spoenemann:
:+1:

Issue #324, opened 2019-06-10
Omitting worspace/symbol query param raises error in language server


jonas-jonas:
When sending `workspace/symbol` requests without a `query` string, LSP4J omits the `query` param altogether. This results in an error being raised in the Dart language server we are using (in combination with LSP4E). 
There is currently a discussion about if the string can be empty or not: https://github.com/Microsoft/language-server-protocol/issues/740. But according to the spec the field should still be sent even if it's empty:
```
{"jsonrpc":"2.0","id":"9","method":"workspace/symbol","params":{}}
```
This is how it should be:
```
{"jsonrpc":"2.0","id":"9","method":"workspace/symbol","params":{"query": ""}}
```

I first raised the issue with the developers of the Dart language server, but according to them it's something that should be fixed in the client (LSP4J here): https://github.com/dart-lang/sdk/issues/37206

jonas-jonas:
See #325

PR #325, opened 2019-06-10
Fixed eclipse#324 Send empty string in query of workspace/symbol

jonas-jonas:
As mentioned in #324 this would fix the error by sending an empty string in the `query` param of `workspace/symbol`. I also locally tested it against our project.

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -4007,6 +4007,7 @@ class WorkspaceSymbolParams {
 	String query
 
 	new() {
+		this.query = '';
 	}
 
 	new(@NonNull String query) {
--- a/org.eclipse.lsp4j/src/main/xtend-gen/org/eclipse/lsp4j/WorkspaceSymbolParams.java
+++ b/org.eclipse.lsp4j/src/main/xtend-gen/org/eclipse/lsp4j/WorkspaceSymbolParams.java
@@ -28,6 +28,7 @@ public class WorkspaceSymbolParams {
   private String query;
   
   public WorkspaceSymbolParams() {
+    this.query = "";
   }
   
   public WorkspaceSymbolParams(@NonNull final String query) {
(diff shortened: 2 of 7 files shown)


spoenemann:
I would prefer to have this fixed in the client implementation (LSP4E) instead of inducing a specific behavior in the default constructor.

jonas-jonas:
While I agree, that this could be fixed in LSP4E, the fix would be passing an empty String to the constructor of `WorspaceSymbolParams`. Which seems kind of "hacky". 

Also the spec states that the `query` should not be empty: https://microsoft.github.io/language-server-protocol/specification#workspace_symbol
(which is currently being discussed: https://github.com/microsoft/vscode-languageserver-node/issues/458)

But in any case the `query` should not be omitted completely. Which means that the implementation in LSP4J is not according to the spec.

That being said, as I mentioned the fix in LSP4E would be rather trivial but "hacky". I just wanted to raise this here first.

spoenemann:
> Which means that the implementation in LSP4J is not according to the spec.

You cannot expect to obtain a valid protocol object just by invoking the default constructor. The default constructor is rather meant for post-construction configuration, which can be done quite nicely in some Java-based languages. For example, in Xtend I'd rather write

```xtend
new WorkspaceSymbolParams => [
    query = "foo"
]
```

instead of using a constructor `new WorkspaceSymbolParams("foo")` because the constructor doesn't tell me to which properties its arguments are assigned. The Xtend syntax is closer to the TypeScript style object creation:

```typescript
<WorkspaceSymbolParams> {
    query: "foo"
}
```

Anyway, I'm not totally against initializing properties in the default constructor (this is already done in a few protocol classes, but it's very inconsistent... maybe worth a separate issue?), but I also don't think it would be "hacky" to call the constructor (or the `query` setter) with an empty string in the LSP4E client. In the pull request provided in the discussion you mentioned

https://github.com/microsoft/language-server-protocol/pull/770/files#diff-32fe5a6c631807387f43b01824ccea9d

the documentation is changed to

> A query string to filter symbols by. Clients may send an empty string here to request all symbols.

I understand this as the empty string having a special meaning in the protocol, so I think it's good if the client explicitly sets the property to `""` to show it's referring to that special meaning.

jonas-jonas:
Alright, makes sense. Thanks for your quick and detailed answers. I will raise the issue with LSP4E. Closing.

PR #343, opened 2019-08-01
Allow to globally disable null-checks of LSP data structures

spoenemann:
Follow-up of #305. This allows to globally disable the null-checks in applications where they cause more trouble than they actually help. See the discussion:
https://github.com/eclipse/xtext-core/pull/1060/files/853bfb6536c406f26ca53f17eef9189e649beb2e#pullrequestreview-210108691

@kittaakos @cdietrich would this change improve the situation?


kittaakos:
> would this change improve the situation?

Thank you for this patch, @spoenemann! I am no sure if I ever want to change the runtime behavior with a boolean flag, but I do not have anything against this PR.

PR #353, opened 2019-08-09
Update to DAP 1.35

KamasamaK:
The supported DAP is very outdated. I've gone through and made all of the necessary changes to bring it to the latest version v1.35.

cc @jonahgraham


jonahgraham:
Why close and delete?

KamasamaK:
@jonahgraham Sorry, I will resubmit. I did not use the proper author information.

PR #364, opened 2019-09-07
Fix prepareRename return type and add tests

### LINKED ISSUE #363: Incorrect return type for textDocument/prepareRename
According to the [language server protocol specification](https://microsoft.github.io/language-server-protocol/specification#textDocument_prepareRename), `textDocument/prepareRename` should return `Range | { range: Range, placeholder: string } | null`.

However, if you look at the relevant adapter implementation, you'll notice that a Collection adapter is being used: https://github.com/eclipse/lsp4j/blob/1db24f59ab63988bda5644a44ba4e93e6c5c3e1c/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/PrepareRenameResponseAdapter.java#L42

This seems incorrect. I spotted this bug by getting `java.util.ArrayList cannot be cast to org.eclipse.lsp4j.jsonrpc.messages.Either` whilst working on the LSP4E project ([Bug 525411](https://bugs.eclipse.org/bugs/show_bug.cgi?id=525411)). I'll add some tests and submit a fix shortly.

PyvesB:
Fixes #363.

Signed-off-by: Pierre-Yves B. <PyvesDev@gmail.com>

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/PrepareRenameResponseAdapter.java
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/PrepareRenameResponseAdapter.java
@@ -1,5 +1,5 @@
 /******************************************************************************
- * Copyright (c) 2019 Microsoft Corporation and others.
+ * Copyright (c) 2018-2019 Microsoft Corporation and others.
  * 
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/PrepareRenameResponseAdapter.java
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/PrepareRenameResponseAdapter.java
@@ -1,5 +1,5 @@
 /******************************************************************************
- * Copyright (c) 2018 Microsoft Corporation and others.
+ * Copyright (c) 2019 Microsoft Corporation and others.
  * 
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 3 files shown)


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/PrepareRenameResponseAdapter.java
```diff
@@ -1,5 +1,5 @@
 /******************************************************************************
- * Copyright (c) 2018 Microsoft Corporation and others.
+ * Copyright (c) 2019 Microsoft Corporation and others.
```
spoenemann: If you update the year in the file header, it should be a range, e.g. `2018-2019`.

https://www.eclipse.org/projects/handbook/#ip-copyright-headers

> the first year is when the content was first created and the last year is when the content was last modified.

org.eclipse.lsp4j/src/test/java/org/eclipse/lsp4j/test/services/JsonParseTest.xtend
```diff
@@ -1,5 +1,5 @@
 /******************************************************************************
- * Copyright (c) 2016 TypeFox and others.
+ * Copyright (c) 2019 TypeFox and others.
```
spoenemann: See above.

org.eclipse.lsp4j/src/test/java/org/eclipse/lsp4j/test/services/MessageMethods.xtend
```diff
@@ -1,5 +1,5 @@
 /******************************************************************************
- * Copyright (c) 2016 TypeFox and others.
+ * Copyright (c) 2019 TypeFox and others.
```
spoenemann: See above.

PyvesB:
@spoenemann copyright headers fixed!

spoenemann:
ok to test

PR #366, opened 2019-09-09
Add Diagnostic Tag Capability

akaroml:
Signed-off-by: Rome Li <rome.li@microsoft.com>

Change from protocol: https://github.com/microsoft/vscode-languageserver-node/blob/fda16d6b63ba0fbdbd21d437ea810685528a0018/protocol/src/protocol.ts#L1313

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -816,12 +816,19 @@ class PublishDiagnosticsCapabilities {
 	 * Whether the client accepts diagnostics with related information.
 	 */
 	Boolean relatedInformation
+	
+	Boolean tagSupport
 
 	new() {
 	}
 
 	new(Boolean relatedInformation) {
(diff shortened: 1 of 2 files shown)


akaroml:
@spoenemann @yaohaizh could you guys take a look?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -816,13 +816,20 @@ class PublishDiagnosticsCapabilities {
 	 * Whether the client accepts diagnostics with related information.
 	 */
 	Boolean relatedInformation
+	
+	Boolean tagSupport
```
spoenemann: Please add the comment as written in the vscode source:
```
	/**
	 * Client supports the tag property to provide meta data about a diagnostic.
	 *
	 * Since 3.15
	 */
```

akaroml: Added.

spoenemann:
Thanks

PR #368, opened 2019-09-13
Added partial message support for WebSocketMessageHandler

jsteuer:
I'm using your great `WebSocketLauncherBuilder` to serve my language server and some additonal endpoints. On large payloads (> 8kB) I'm running into trouble:

> The decoded text message was too big for the output buffer and the endpoint does not support partial messages

To fix that, I changed the class `WebSocketMessageHandler` to support partial messages.

If your not happy with that, it would be great, if you provide some other 'official' way to support larger payloads. 

Thank you.

---

**Note**: The `dispatch` method of the `MockSession` makes things very smart since handlers implementing `MessageHandler.Whole` finally getting active - even on fragmented messages. In reason of that, the unit test I added, would not break using the old `MessageHandler.Whole` handler. It seems, that some implementations like Tomcat are less smart. They will only call `MessageHandler.Partial` handlers - and never `MessageHandler.Whole` handlers - for fragmented messages. See [here](https://github.com/apache/tomcat/blob/master/java/org/apache/tomcat/websocket/WsFrameBase.java), specialy the places where `"wsFrame.textMessageTooBig"` - the key to get the error message above - is used.

### DIFF
--- a/org.eclipse.lsp4j.websocket/src/main/java/org/eclipse/lsp4j/websocket/WebSocketMessageHandler.java
+++ b/org.eclipse.lsp4j.websocket/src/main/java/org/eclipse/lsp4j/websocket/WebSocketMessageHandler.java
@@ -22,19 +22,41 @@ import org.eclipse.lsp4j.jsonrpc.messages.Message;
 /**
  * WebSocket message handler that parses JSON messages and forwards them to a {@link MessageConsumer}.
  */
-public class WebSocketMessageHandler implements MessageHandler.Whole<String> {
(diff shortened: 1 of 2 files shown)


spoenemann:
The JavaDoc of `MessageHandler.Whole` says:

> This kind of handler is notified by the container on arrival of a complete message. If the message is received in parts, the container buffers it until it is has been fully received before this method is called.

It seems that those implementations that are "less smart" do not implement the websocket API correctly?

jsteuer:
You're absolutely right! So this pull request is nonsense. I do not know how I could missed out this JavaDoc...

PR #376, opened 2019-10-29
Update LSP4J to DAP 1.37.0

jonahgraham:
This PR bring DAP support to 1.37.0. However I have not implemented the cancel request added in 1.37.0 because I think that needs further thought on how to tie in to LSP4J's built-in cancellation support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### v0.9.0 (TBD)
 
  * Implemented DAP version 1.36.0
+ * Implemented DAP version 1.37.0 (except CancelRequest)
 
 Fixed issues: https://github.com/eclipse/lsp4j/milestone/15?closed=1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,30 @@
 ## LSP4J Change Log
 
+### v0.9.0 (TBD)
+
+ * Implemented DAP version 1.36.0
+
+Fixed issues: https://github.com/eclipse/lsp4j/milestone/15?closed=1
+
+Breaking API changes:
+ * In DebugProtocol many fields which used to be defined as `Long` are now `Integer`.
+   This is due to changes in DAP 1.36 that clarified these values should be less
+   than or equal to 2147483647 (2^31 - 1)
+    - `OutputEventArguments`: `variablesReference`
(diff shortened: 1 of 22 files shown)


jonahgraham:
@KamasamaK Would you like to have a look at this update and have a review?

jonahgraham:
> but at a glance it looks like some of the Long variables should actually be Integer now.

I had it that way in my first pass and then changed my mind -- it is actually one of the things I was hoping to solicit feedback on. Changing this would change the API in a breaking way - but that may be ok. It is fine for LSP4E to make this change.

jonahgraham:
I'm going to send an email to lsp4j-dev to solicit feedback.

KamasamaK:
Yes, I was considering that would be breaking too, but I think it's better to force compliance with the spec than to leave it in a comment. Also, I don't know what the approach is to versioning with eclipse projects, but I expect pre-1.0 software may have breaking changes.

jonahgraham:
@KamasamaK  - if you have time and want to re-review I have changed and documented all the Longs that change to Integers. 

Note that there are other fields which are still Long, and it would be worth someone to reach out to DAP folk for clarity, but for now I have only changed those fields explicitly documented as fitting in Integers.

KamasamaK:
@jonahgraham What was the reason for keeping some of the fields as `Long`? All of the numeric properties in the [DAP schema](https://github.com/microsoft/debug-adapter-protocol/blob/gh-pages/debugAdapterProtocol.json) appear to be `integer` now. Unfortunately, it doesn't appear that the auto-generated spec file accounts for this since it just uses TypeScript representations. I've opened [an issue with DAP](https://github.com/microsoft/debug-adapter-protocol/issues/90) for this.

jonahgraham:
Those items, such as threadId have always been integer so have always been mapped to 64-bit. IIRC when I first wrote this the Python debug adapter had threadIds that exceeded 32-bits. 

I have no objection to changing them and I am grateful for you raising the issue with DAP. Thanks!

jonahgraham:
Yes, as it turns out the >32-bit in Python was actually a bug: https://github.com/microsoft/vscode-python/pull/761

Reading through the history that you have shared really indicates that it should have always been 32-bits always. I'll create a new issue.

Issue #369, opened 2019-09-17
Release v0.9.0


mickaelistria:
In LSP4E, we'd like to take advantage of some new features of LSP4J, for example to enable https://bugs.eclipse.org/bugs/show_bug.cgi?id=525411 . Could a release happen soon?

mickaelistria:
cc @PyvesB

spoenemann:
We could make 0.8.1 based on the current master state? Version 0.9.0 would require a release review and thus a delay of several weeks.

cdietrich:
@spoenemann release reviews are usually required once a year recently

spoenemann:
Is it documented somewhere? I haven't found anything in https://www.eclipse.org/projects/handbook/#release-review
CC @waynebeaton

cdietrich:
@spoenemann the same document states

> A Project can make official Releases for one calendar year following a successful Progress Review or Release Review. The Project Leadership Chain may—​at its discretion—​require that the Project engage in additional Reviews (e.g. a progress or Release Review) prior to issuing a Release.

mickaelistria:
Both a 0.8.1 or a 0.9.0 would be OK with me.
But @cdietrich is right, although the PMI doesn't make it clear, projects are now free to release minor versions just like the do for micro versions, without delay. A review (and related delay) is only required once a year now.

waynebeaton:
Making the PMI more helpful is on my list.

Service releases that are based on a previous major or minor release require no review (with no time restrictions). Service releases are releases that do not contain any new functionality (e.g., bug fixes only). What number you use is irrelevant: when a release includes new functionality, it is by definition either a major or minor release.

spoenemann:
@mickaelistria version 0.8.1 is on its way with the prepareRename fix. For 0.9.0 I'd like to also include at least #370.

KamasamaK:
Ideally, everything introduced in LSP 3.15 should make LSP4J 0.9.0. You can see a preview of the upcoming version at https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/.

mickaelistria:
> Ideally, everything introduced in LSP 3.15 should make LSP4J 0.9.0.

I guess it's up to those who need a feature or another to help make this happen.
[...]

PR #378, opened 2019-11-26
Bug #369: Add signing and deploying for lsp4j

jonahgraham:
Prior to this the git repo that handled signing was outside
of Eclipse.


README.md
```diff
@@ -18,16 +18,13 @@ A brief overview of how to use LSP4J to implement a server or a client can be fo
...
-The snapshots are regularly deployed to:
+The snapshots are deployed during the build to:
  * https://oss.sonatype.org/content/repositories/snapshots/org/eclipse/lsp4j/
 
-Also, you can get them directly from the TypeFox Jenkins server:
- * http://services.typefox.io/open-source/jenkins/job/lsp4j/job/master/lastSuccessfulBuild/artifact/build/maven-repository/
-
 ### Supported LSP Versions
 
  * LSP4J 0.7.&ast;, 0.8.&ast; &rarr; LSP 3.14.0
```
cdietrich: list misses 0.9

releng/build.Jenkinsfile
```diff
@@ -0,0 +1,59 @@
+pipeline {
+  agent any
```
cdietrich: sure this will work on any agent?

releng/pom.xml
```diff
@@ -12,6 +12,12 @@
 		<root-dir>${basedir}/..</root-dir>
 	</properties>
 
+	<pluginRepositories>
```
cdietrich: calling `./gradlew generateP2Build` will move these changes to the void again unfortunately

cdietrich: https://github.com/TypeFox/gradle-plugins/blob/9742b36882f062d272917528c5f5ccfec3ee21a5/gradle-p2gen/src/main/java/io/typefox/gradle/p2gen/P2GenPlugin.xtend

cdietrich: => we would need to fix that there or get rid of the p2gen an maintain the stuff manually as we did in Xtext

releng/release-eclipse.sh
```diff
@@ -0,0 +1,46 @@
...
+wget -q --output-document=/dev/null $ARTIFACTS_REPO_TARGET
+
+ECHO=echo
+if [ "$DRY_RUN" == "false" ]; then
+    ECHO=""
+else
+    echo Dry run of build:
+fi
+
+$ECHO mkdir -p $DOWNLOAD_MOUNT
```
cdietrich: this wont do a drops deployment correct?
seems like it was not done for some releases.
am not sure if this is nice for people who want to download the repo
https://download.eclipse.org/lsp4j/downloads/drops/?d

gradle/java-compiler-settings.gradle
```diff
@@ -35,6 +35,49 @@ task javadocJar(type: Jar, dependsOn: javadoc) {
 	from javadoc.destinationDir
 }
 
+ext.signMethod = { jarfile -> 
+	println "Signing $jarfile"
+	def SIGNING_SERVICE = 'http://build.eclipse.org:31338/sign'
+	def STDOUT_FORMAT = '    %{size_upload} bytes uploaded, %{size_download} bytes downloaded (%{time_total} s)\\n'
+
+	ProcessBuilder curl_pb = new ProcessBuilder("curl", 
+		"--fail", "--silent", "--show-error", "--output",
```
cdietrich: we had to add some retrys to Xtext here as the service is instable sometimes

releng/pom.xml
jonahgraham: OK. I hadn't realized it was being generated (TODO add "this file is auto-generated" comment)

jonahgraham: For xtext did you just start with the generated file and remove the generateP2Build?

gradle/java-compiler-settings.gradle
cdietrich: https://git.eclipse.org/c/cbi/org.eclipse.cbi.git/tree/maven-plugins/eclipse-jarsigner-plugin/src/main/java/org/eclipse/cbi/maven/plugins/jarsigner/mojo/SignMojo.java does 3 retrys too

releng/release-eclipse.sh
jonahgraham: The "drops" is a separate issue and if someone needs it they can request and I will look into doing that. Do you need it @cdietrich?

releng/pom.xml
cdietrich: yes.

cdietrich: i am not sure if we can change the typefox plugin to add the repo and profile

gradle/java-compiler-settings.gradle
jonahgraham: Thanks, I had seen some retries in other code - I can look at that once we have some instability - at the moment the full build time is very short so running it again is less time consuming. I assume xtext has a much longer build time, so having a signing failure would be much more inconvenient?

releng/release-eclipse.sh
cdietrich: not right now

releng/build.Jenkinsfile
jonahgraham: Any agent that lsp4j CI currently has. If (when!) it is moved to JIRO more will be needed here.

README.md
jonahgraham: Fixed.

gradle/java-compiler-settings.gradle
cdietrich: indeed

cdietrich:
@jonahgraham will this do deployments for PRs too or for branches only?

releng/pom.xml
jonahgraham: I have changed this to be manually generated and removed generateP2Build.

cdietrich: ok. can you please  create a follow up task to discuss this with @spoenemann or @dhuebner  once they are back?

jonahgraham:
> @jonahgraham will this do deployments for PRs too or for branches only?

It will do deploy to download.eclipse.org for branches in this repo - but it will not deploy to maven central except for master and release_* named branches. See https://github.com/eclipse/lsp4j/pull/378/files#diff-c89773d0f8e78cce85e72b37c626d0a7R23

gradle/p2-deployment.gradle
```diff
@@ -1,53 +0,0 @@
...
- * This program and the accompanying materials are made available under the
- * terms of the Eclipse Public License v. 2.0 which is available at
- * http://www.eclipse.org/legal/epl-2.0,
- * or the Eclipse Distribution License v. 1.0 which is available at
- * http://www.eclipse.org/org/documents/edl-v10.php.
- * 
- * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
- ******************************************************************************/
-
-apply plugin: 'io.typefox.gradle.p2gen'
```
cdietrich: plugin is still loaded in build.gradle

cdietrich:
ok. we might need a follow up task for removal of branches later

gradle/p2-deployment.gradle
jonahgraham: I think I removed it: https://github.com/eclipse/lsp4j/pull/378/files#diff-c197962302397baf3a4cc36463dce5eaL27

cdietrich: i mean this one https://github.com/eclipse/lsp4j/blob/ec16cc8b2dfa797085be596fc6dd743ea5a4bc65/build.gradle#L20

jonahgraham: Removed too: https://github.com/eclipse/lsp4j/pull/378/files#diff-c197962302397baf3a4cc36463dce5eaL20

jonahgraham: PS I don't know how you made your link to the source code look pretty (with inlined source) and mine are ugly links.

cdietrich:
ok. in general this looks ok to me.
i still have no clue who maintains
e.g.
http://download.eclipse.org/lsp4j/updates/releases/compositeContent.jar
http://download.eclipse.org/lsp4j/updates/releases/compositeArtifacts.jar

gradle/p2-deployment.gradle
cdietrich: i used the rightklick copy permalink

jonahgraham:
> i still have no clue who maintains

Personally I would propose removing them. I can't understand in what circumstances a consumer would correctly want that composite. There is an argument for having composites that cover 0.8.0 and 0.8.1 in the same composite though, but that is not what is in there.

cdietrich:
the only usecase is basically: always take the latest and hope the best.
in Xtext we have a composite forever. and we have adapted the jobs to adapt them
but we dont have jars but xmls.
=> maybe we should leave it for now and create another follow up tasks
when we know how it was created in the past we might be able to replicate
it.

gradle/p2-deployment.gradle
jonahgraham: Thanks - that isn't an option in the diff viewer, but for completeness on this resolved conversation here is the link in the new file showing it gone:

https://github.com/eclipse/lsp4j/blob/fe5212dd8a2a7bed66cc30c669a9aac58226ef34/build.gradle#L18-L20

PR #387, opened 2019-12-13
Update call hierarchy for lsp4j

testforstephen:
The call hierarchy protocol in LSP has updated to a new version, see https://github.com/microsoft/vscode-languageserver-node/blob/master/protocol/src/protocol.callHierarchy.proposed.ts.

It includes three new requests:
- textDocument/prepareCallHierarchy
- callHierarchy/incomingCalls
- callHierarchy/outgoingCalls

Signed-off-by: Jinbo Wang <jinbwan@microsoft.com>


jonahgraham:
> Can one of the admins verify this patch?

run tests

(Note I kicked off the tests, but I am not in a position to review at the moment)

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -4850,73 +4850,102 @@ class SemanticHighlightingInformation {
...
+	CallHierarchyItem from
 
 	/**
-	 * The symbol that contains the reference.
+	 * The range at which at which the calls appears. This is relative to the caller
+	 * denoted by [`this.from`](#CallHierarchyIncomingCall.from).
 	 */
-	@NonNull
-	CallHierarchySymbol from
+	Range[] fromRanges
```
KamasamaK:
```suggestion
	List<Range> fromRanges
```

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -4850,73 +4850,102 @@ class SemanticHighlightingInformation {
...
-	@NonNull
-	CallHierarchySymbol to
+	CallHierarchyItem to
 
+	/**
+	 * The range at which this item is called. This is the range relative to the caller, e.g the item
+	 * passed to [`provideCallHierarchyOutgoingCalls`](#CallHierarchyItemProvider.provideCallHierarchyOutgoingCalls)
+	 * and not [`this.to`](#CallHierarchyOutgoingCall.to).
+	 */
+	Range[] fromRanges
```
KamasamaK:
```suggestion
	List<Range> fromRanges
```

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -4850,73 +4850,102 @@ class SemanticHighlightingInformation {
...
+	 * The kind of this item.
 	 */
 	@NonNull
 	SymbolKind kind
 
 	/**
-	 * URI of the document containing the symbol.
+	 * Tags for this item.
+	 */
+	SymbolTag[] tags
```
KamasamaK:
```suggestion
	List<SymbolTag> tags
```

KamasamaK: `SymbolTag` is not defined here.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
testforstephen: fixed

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
testforstephen: SymbolTag is enum type. I saw other enums are all manually written. Is there any enum sample for Xtend definition?

testforstephen: Looks like enum with constructor is not supported well in Xtend. https://bugs.eclipse.org/bugs/show_bug.cgi?id=403783

KamasamaK: You're right, the enums for LSP seem to be manual. I must have just remembered seeing them in [DebugProtocol.xtend](https://github.com/eclipse/lsp4j/blob/master/org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/DebugProtocol.xtend).

testforstephen: Yeah, i have a try on adding SymbolTag enum to Xtend, but it doesn't work.

Here are the problems:
- SymbolTag in LSP starts with 1, so the default enum definition doesn't work.
```java
enum SymbolTag {
  Deprecated
}
```
- @SerializedName can only specify an alias string for the enum type. But the value passed from LSP is a int number, the jsonrpc cannot convert it to the serializedName SymbolTag.
```java
enum SymbolTag {
  @SerializedName("1")
  Deprecated
}
```

Issue #392, opened 2020-01-03
DAP has integer in the spec that probably should be mapped to Integer instead of Long


jonahgraham:
@jonahgraham What was the reason for keeping some of the fields as `Long`? All of the numeric properties in the [DAP schema](https://github.com/microsoft/debug-adapter-protocol/blob/gh-pages/debugAdapterProtocol.json) appear to be `integer` now. Unfortunately, it doesn't appear that the auto-generated spec file accounts for this since it just uses TypeScript representations. I've opened [an issue with DAP](https://github.com/microsoft/debug-adapter-protocol/issues/90) for this.

_Originally posted by @KamasamaK in https://github.com/eclipse/lsp4j/pull/376#issuecomment-570404313_

jonahgraham:
@KamasamaK I have put this to be handled before 0.9.0 ships so that we are only changing this API once.

jonahgraham:
Fixed merged as part of PR #401

PR #401, opened 2020-01-27
Fix #150 and #392 - remove use of Long and use primitives on @NonNull fields

### LINKED ISSUE #150: Use primitive types in debug protocol
For LSP we are using primitive types for required fields where applicable. This could also be done for the debug protocol. For example,

```
	@NonNull
	Long threadId;
```

could be changed to

```
	long threadId;
```

I'm not sure about the `@NonNull` annotation in that case. We have omitted it for LSP primitive fields, but it could still be useful to enable validation for those fields, so we might change that.

jonahgraham:


jonahgraham:
@KamasamaK you can have a look if you have time. I will try rolling this out for the M2 date next week, but we still have over a month until the 0.9.0 release is scheduled.

KamasamaK:
It's not letting me comment on these lines, but there are two `@NonNull Boolean` fields that were missed for replacing with primitive. Aside from that, LGTM.

Issue #407, opened 2020-02-18
Incompatibility with vscode-languageclient < 6.0.0


JanKoehnlein:
VS Code initially added the `PublishDiagnosticsCapabilities#tagSupport` as boolean in the `vscode-language-client` implementation, but without adding it to the LSP spec. In LSP 3.15 it was finally speced out as being of type `DiagnosticsTagSupport` and the client implementation was adapted in version 6.0.0.

So all clients relying on `vscode-language-client` < 6.0.0 will fail to connect to a lsp4j server as it cannot parse their initialize request. Unfortunately, that includes the `monaco-language-client`, and thereby all extensions to Eclipse Theia. It's unlikely this can be fixed within the next weeks.

For the time being, I'd propose to change the type of `PublishDiagnosticsCapabilities#tagSupport` to `Either<Boolean,  DiagnosticsTagSupport>`.

PR #408, opened 2020-02-18
[compatibility] improve compatibility with vscode-languageserver < 6.0.0

JanKoehnlein:
Fixes eclipse/lsp4j#407


jonahgraham:
@eclipse-lsp4j-bot  ok to test

@JanKoehnlein Although this doesn't strictly break API because there hasn't been a release with tagSupport yet I think this is a candidate that could still be mentioned in the [changelog](https://github.com/eclipse/lsp4j/blob/master/CHANGELOG.md) as I imagine the problem is people are using non-release versions of LSP4J to have encountered this problem?

Tagging @testforstephen and @akaroml because they contributed the earlier versions of tagSupport.

PR #409, opened 2020-02-19
Remove @Beta from selectionRangeProvider

KamasamaK:
When I was removing the beta tags for selection range, I seem to have overlooked this one. I also added some missing "Since" doc strings while I was looking over what else might have been missed.


jonahgraham:
@KamasamaK  Is this ready to go? If you don't have your commit rights setup yet I can do the merge.

KamasamaK:
@jonahgraham Yes, it's ready. I cannot merge.

PR #412, opened 2020-02-19
Added comprehensive request params

### LINKED ISSUE #410: Comprehensive request params
As part of LSP 3.15, all requests now have explicitly defined request params. The motivation behind this was for supporting the new progress features, which are not yet implemented for LSP4J, but I was thinking it might be good to get this out of the way sooner than later. One reason for this is to prevent future compatibility breaks like what happened with `signatureHelp` in #397. This would unfortunately result in several compatibility breaks now, but it is my opinion that it's better to get it done all at once and prior to 1.0. I've compiled the following request params that would need to be created and used instead of the current `TextDocumentPositionParams`:

- HoverParams
- DeclarationParams
- DefinitionParams
- TypeDefinitionParams
- ImplementationParams
- DocumentHighlightParams

For now, they would just extend `TextDocumentPositionParams` with no additional fields.

I can create the PR for this if we're in agreement to go forward.

KamasamaK:
Fixes #410

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## LSP4J Change Log
 
-### v0.9.0 (TBD)
+### v0.9.0 (Feb. 2020)
 
  * Implemented DAP version 1.36.0
  * Implemented DAP version 1.37.0 (except CancelRequest)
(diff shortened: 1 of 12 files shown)


jonahgraham:
@KamasamaK  Is this ready to go? If you don't have your commit rights setup yet I can do the merge.

KamasamaK:
@jonahgraham Yes, it's ready. I cannot merge, but even if I could I was waiting on potential review comments.

cdietrich:
Rested it and looks fine

Issue #413, opened 2020-02-26
The spec allows the Diagnostic code to be a string or a number.


jonahgraham:
PR#388 says "The spec allows the Diagnostic code to be a string or a number." Unfortunately the submitter cannot contribute code to LSP4J due to an inability to sign the contributor agreement.

jonahgraham:
@KamasamaK Any chance you can look at this location where the spec and LSP4J differ and recommend whether to fix it/change it? As this is likely yet another API change, it seems getting it in before 0.9.0 would be good.

KamasamaK:
I would recommend fixing since:

1. We should be spec-compliant wherever possible.
2. Wouldn't leaving it break clients using LSP4J if the server sends a numeric code?

cdietrich:
Will this be done or not? Is there a rebased branch available I can test with

jonahgraham:
Someone needs to fix this code to meet spec. I will do it later today if no one else has by then. Unfortunately, despite a submission in pr#388, the contributor can't/won't sign eca.

jonahgraham:
I'm doing this now

PR #417, opened 2020-02-28
Bug #413: Allows the Diagnostic code to be a string or a number

jonahgraham:
This matches https://microsoft.github.io/language-server-protocol/specifications/specification-current/#diagnostic


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -1678,7 +1678,7 @@ class Diagnostic {
 	/**
 	 * The diagnostic's code. Can be omitted.
 	 */
-	String code
+	Either<String, Number> code
```
cdietrich: do we have any generate criteria when we turn number into Integer?

jonahgraham:
Closing in favour of #418 which gives a p2 repo on download.eclipse.org to test against.

PR #418, opened 2020-02-28
Bug #413: Allows the Diagnostic code to be a string or a number

jonahgraham:
This matches https://microsoft.github.io/language-server-protocol/specifications/specification-current/#diagnostic

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -1678,7 +1678,7 @@ class Diagnostic {
 	/**
 	 * The diagnostic's code. Can be omitted.
 	 */
-	String code
+	Either<String, Number> code
 
 	/**
 	 * A human-readable string describing the source of this diagnostic, e.g. 'typescript' or 'super lint'.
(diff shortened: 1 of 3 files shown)


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -1678,7 +1678,7 @@ class Diagnostic {
 	/**
 	 * The diagnostic's code. Can be omitted.
 	 */
-	String code
+	Either<String, Number> code
```
jonahgraham: @cdietrich said:

> do we have any general criteria when we turn number into Integer?

cdietrich:
i had integrated the branch into Xtext and with a few adaptations it works again

jonahgraham:
https://download.eclipse.org/lsp4j/builds/jonah/bug_413/ is p2 repo with this change.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
cdietrich: also: should we offer a "comfort-constructor" with Integer or Long or Number?

cdietrich:
Anything that speaks against bringing this onto master and releasing ?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
jonahgraham: Yes, we should - as adding that does not break API it can be added more easily and as I am not sure what is best needed here I will leave the general case for now.

PR #421, opened 2020-03-21
Deprecate deprecated

ckipp01:
According to https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_completion the `deprecated` field in `CompletionItem` should be deprecated in favor of using `tags`.


KamasamaK:
Please see the [Contribution Guide](https://github.com/eclipse/lsp4j/blob/master/Contributing.md).

You need to make your changes here:
https://github.com/eclipse/lsp4j/blob/d9331c8d74f9ec328da27e8f28bddcb6c230efba/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend#L1524-L1527

PR #427, opened 2020-05-20
TextDocumentSyncOptions.save is either Boolean or SaveOptions

### LINKED ISSUE #426: LSP4J fails to parse TextDocumentSyncOptions with a boolean value for property `save`
The type `TextDocumentSyncOptions` is defined as follows in the LSP spec:

```typescript
export interface TextDocumentSyncOptions {
	...
	save?: boolean | SaveOptions;
}
```

However, LSP4J doesn't handle boolean values, only `SaveOptions`.

Note that this may stem from a slight inconsistency in the LSP spec, which got resolved only quite recently. See also https://github.com/microsoft/language-server-protocol/issues/989

planger:
Fixes #426

Signed-off-by: Philip Langer <planger@eclipsesource.com>

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -2225,7 +2225,7 @@ class TextDocumentSyncOptions {
 	/**
 	 * Save notifications are sent to the server.
 	 */
-	SaveOptions save
+	Either<Boolean, SaveOptions> save
 }
 
 /**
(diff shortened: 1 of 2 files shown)


jonahgraham:
@eclipse-lsp4j-bot run tests

aarcangeli:
Can we merge? I'm having this problem too.
Is there a temporary workaround?

planger:
> Can we merge? I'm having this problem too.

There is a discussion on how to deal with API breaks [1]. As far as I can see, there is some feedback to be in favor of accepting API breaks between minor versions, iff they only reflect LSP spec changes. So I guess this would mean we could merge this change. @jonahgraham what do you think?

> Is there a temporary workaround?

No, none that I'm aware of. I worked around it only by patching LSP4J as a temporary solution.

[1] https://www.eclipse.org/lists/lsp4j-dev/msg00049.html

jonahgraham:
I think we should merge this change and schedule a release. There are a number of outstanding items waiting for a release.

cdietrich:
we also might need a review for the release as the last is >12 month ago

aarcangeli:
Thank you, I just built by myself jars the from master branch and everything is working fine.

PR #431, opened 2020-06-27
Migrate utils classes to lsp4j Preconditions where it is possible

nixel2007:


spoenemann (approved):
Thanks for the contribution, and sorry for the late review!

PR #437, opened 2020-08-14
fix #396 : add missing constructor

### LINKED ISSUE #396: Hover class has no constructor for single MarkedString as contents
It seems this is because the type of the `contents` field in `Hover` was changed from `Either<T, List<T>>` to `List<T>` in https://github.com/eclipse/lsp4j/commit/6ff370566cc40af6aeaaffe8f9c79ac00e8641a3, but this also removed the constructor. Perhaps this should be added back for spec compliance?

Dufgui:


Dufgui:
ready for review

spoenemann (approved):
Thanks!

PR #438, opened 2020-08-14
fix #422 align deprecated on completionItem

### LINKED ISSUE #422: Deprecate deprecated in CompletionItem
According to https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_completion the `deprecated` field in `CompletionItem` should be deprecated in favor of using `tags`.

Dufgui:


Dufgui:
ready for review

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -1523,7 +1523,9 @@ class CompletionItem {
 
 	/**
 	 * Indicates if this item is deprecated.
+	 * @deprecated Use `tags` instead if supported.
```
KamasamaK, line 1526: Separate with a new line to be consistent with other `@deprecated` Javadoc tags.

KamasamaK (approved):
LGTM

PR #442, opened 2020-08-18
fix #370 add progress management

### LINKED ISSUE #370: Progress support in LSP
LSP seems supporting progress support in LSP: 

* https://github.com/microsoft/language-server-protocol/issues/786
* https://github.com/Microsoft/vscode-languageserver-node/blob/master/protocol/src/protocol.progress.proposed.ts#L1

It should be fantastic if LSP4J could support it. Thank's!

Dufgui:
add the progress  management

### DIFF
--- a/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/EitherTypeAdapter.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/EitherTypeAdapter.java
@@ -201,16 +201,14 @@ public class EitherTypeAdapter<L, R> extends TypeAdapter<Either<L, R>> {
 	protected Either<L, R> createLeft(L obj) throws IOException {
 		if (Either3.class.isAssignableFrom(typeToken.getRawType()))
 			return (Either<L, R>) Either3.forLeft3(obj);
-		else
-			return Either.forLeft(obj);
+		return Either.forLeft(obj);
 	}
 	
 	@SuppressWarnings("unchecked")
 	protected Either<L, R> createRight(R obj) throws IOException {
 		if (Either3.class.isAssignableFrom(typeToken.getRawType()))
 			return (Either<L, R>) Either3.forRight3((Either<?, ?>) obj);
-		else
-			return Either.forRight(obj);
+		return Either.forRight(obj);
 	}
 
 	protected static class EitherTypeArgument<T> {
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -24,6 +24,7 @@ import org.eclipse.lsp4j.adapters.ResourceChangeListAdapter
 import org.eclipse.lsp4j.adapters.ResourceOperationTypeAdapter
 import org.eclipse.lsp4j.adapters.SymbolInformationTypeAdapter
 import org.eclipse.lsp4j.adapters.VersionedTextDocumentIdentifierTypeAdapter
+import org.eclipse.lsp4j.adapters.WorkDoneProgressNotificationAdapter
 import org.eclipse.lsp4j.generator.JsonRpcData
 import org.eclipse.lsp4j.jsonrpc.json.adapters.JsonElementTypeAdapter
 import org.eclipse.lsp4j.jsonrpc.messages.Either
@@ -1184,6 +1185,20 @@ class TextDocumentClientCapabilities {
 }
 
 /**
+ * Window specific client capabilities.
+ */
+@JsonRpcData
+class WindowClientCapabilities {
+	/**
+	 * Whether client supports handling progress notifications. If set servers are allowed to
+	 * report in `workDoneProgress` property in the request specific server capabilities.
+	 *
+	 * Since 3.15.0
+	 */
+	Boolean workDoneProgress;
+}
+
+/**
  * `ClientCapabilities` now define capabilities for dynamic registration, workspace and text document features the client supports.
  * The `experimental` can be used to pass experimental capabilities under development.
  * For future compatibility a `ClientCapabilities` object literal can have more properties set than currently defined.
@@ -1209,6 +1224,11 @@ class ClientCapabilities {
 	TextDocumentClientCapabilities textDocument
 
 	/**
+	 * Window specific client capabilities.
+	 */
+	WindowClientCapabilities window
+
+	/**
 	 * Experimental client capabilities.
 	 */
 	@JsonAdapter(JsonElementTypeAdapter.Factory)
@@ -1220,6 +1240,13 @@ class ClientCapabilities {
 	new(WorkspaceClientCapabilities workspace, TextDocumentClientCapabilities textDocument, Object experimental) {
 		this.workspace = workspace
 		this.textDocument = textDocument
+		this.experimental = experimental
+	}
+	
+	new(WorkspaceClientCapabilities workspace, TextDocumentClientCapabilities textDocument, WindowClientCapabilities window, Object experimental) {
+		this.workspace = workspace
+		this.textDocument = textDocument
+		this.window = window
 		this.experimental = experimental
 	}
 }
@@ -1321,7 +1348,8 @@ class CodeActionContext {
  * These commands are typically code fixes to either fix problems or to beautify/refactor code.
  */
 @JsonRpcData
-class CodeActionParams {
+class CodeActionParams extends WorkDoneProgressAndPartialResultParams {
+
 	/**
 	 * The document in which the command was invoked.
 	 */
@@ -1394,7 +1422,7 @@ class CodeLens {
  * Code Action options.
  */
 @JsonRpcData
-class CodeActionOptions {
+class CodeActionOptions extends AbstractWorkDoneProgressOptions {
 	/**
 	 * CodeActionKinds that this server may return.
 	 *
@@ -1415,7 +1443,7 @@ class CodeActionOptions {
  * Code Lens options.
  */
 @JsonRpcData
-class CodeLensOptions {
+class CodeLensOptions extends AbstractWorkDoneProgressOptions {
 	/**
 	 * Code lens has a resolve provider as well.
 	 */
@@ -1433,7 +1461,8 @@ class CodeLensOptions {
  * The code lens request is sent from the client to the server to compute code lenses for a given text document.
  */
 @JsonRpcData
-class CodeLensParams {
+class CodeLensParams extends WorkDoneProgressAndPartialResultParams {
+
 	/**
 	 * The document to request code lens for.
 	 */
@@ -1638,7 +1667,8 @@ class CompletionList {
  * Completion options.
  */
 @JsonRpcData
-class CompletionOptions {
+class CompletionOptions extends AbstractWorkDoneProgressOptions {
+
 	/**
 	 * The server provides support to resolve additional information for a completion item.
 	 */
@@ -1994,6 +2024,7 @@ class WillSaveTextDocumentParams {
  */
 @JsonRpcData
 class DocumentFormattingParams {
+	
 	/**
 	 * The document to format.
 	 */
@@ -2107,7 +2138,8 @@ class DocumentLink {
  * The document links request is sent from the client to the server to request the location of links in a document.
  */
 @JsonRpcData
-class DocumentLinkParams {
+class DocumentLinkParams extends WorkDoneProgressAndPartialResultParams {
+	
 	/**
 	 * The document to provide document links for.
 	 */
@@ -2144,7 +2176,8 @@ class DocumentLinkOptions {
  * Execute command options.
  */
 @JsonRpcData
-class ExecuteCommandOptions {
+class ExecuteCommandOptions extends AbstractWorkDoneProgressOptions {
+
 	/**
 	 * The commands to be executed on the server
 	 */
@@ -2281,6 +2314,7 @@ class DocumentOnTypeFormattingOptions {
  */
 @JsonRpcData
 class DocumentOnTypeFormattingParams extends DocumentFormattingParams {
+	
 	/**
 	 * The position at which this request was send.
 	 */
@@ -2313,7 +2347,10 @@ class DocumentOnTypeFormattingParams extends DocumentFormattingParams {
  * The document range formatting request is sent from the client to the server to format a given range in a document.
  */
 @JsonRpcData
-class DocumentRangeFormattingParams extends DocumentFormattingParams {
+class DocumentRangeFormattingParams extends DocumentFormattingParams implements WorkDoneProgressParams {
+	
+	Either<String, Number> workDoneToken
+	
 	/**
 	 * The range to format
 	 */
@@ -2390,11 +2427,20 @@ class ResolveTypeHierarchyItemParams {
 
 }
 
+@JsonRpcData
+class DocumentSymbolOptions extends AbstractWorkDoneProgressOptions {
+}
+
+@JsonRpcData
+class DocumentSymbolRegistrationOptions extends AbstractTextDocumentRegistrationAndWorkDoneProgressOptions {
+}
+
 /**
  * The document symbol request is sent from the client to the server to list all symbols found in a given text document.
  */
 @JsonRpcData
-class DocumentSymbolParams {
+class DocumentSymbolParams extends WorkDoneProgressAndPartialResultParams {
+
 	/**
 	 * The text document.
 	 */
@@ -2703,6 +2749,296 @@ class MarkedString {
 	}
 }
 
+/**
+ * The $/progress notification payload interface.
+ *
+ * Since 3.15.0
+ */
+interface WorkDoneProgressNotification {
+	def WorkDoneProgressKind getKind()
+}
+ 
+/**
+ * The $/progress notification payload to start progress reporting.
+ *
+ * Since 3.15.0
+ */
+@JsonRpcData
+class WorkDoneProgressBegin implements WorkDoneProgressNotification {
+
+	/**
+	 * Always return begin
+	 */
+	override WorkDoneProgressKind getKind() {
+		WorkDoneProgressKind.begin
+	}
+
+	/**
+	 * Mandatory title of the progress operation. Used to briefly inform about
+	 * the kind of operation being performed.
+	 *
+	 * Examples: "Indexing" or "Linking dependencies".
+	 */
+	@NonNull
+	String title
+
+	/**
+	 * Controls if a cancel button should show to allow the user to cancel the
+	 * long running operation. Clients that don't support cancellation are allowed
+	 * to ignore the setting.
+	 */
+	Boolean cancellable
+
+	/**
+	 * Optional, more detailed associated progress message. Contains
+	 * complementary information to the `title`.
+	 *
+	 * Examples: "3/25 files", "project/src/module2", "node_modules/some_dep".
+	 * If unset, the previous progress message (if any) is still valid.
+	 */
+	String message
+
+	/**
+	 * Optional progress percentage to display (value 100 is considered 100%).
+	 * If not provided infinite progress is assumed and clients are allowed
+	 * to ignore the `percentage` value in subsequent in report notifications.
+	 *
+	 * The value should be steadily rising. Clients are free to ignore values
+	 * that are not following this rule.
+	 */
+	Integer percentage;
+}
+
+/**
+ * The notification payload about progress reporting.
+ *
+ * Since 3.15.0
+ */
+@JsonRpcData
+class WorkDoneProgressReport implements WorkDoneProgressNotification {
+
+	/**
+	 * Always return report
+	 */
+	override WorkDoneProgressKind getKind() {
+		WorkDoneProgressKind.report
+	}
+
+	/**
+	 * Controls enablement state of a cancel button. This property is only valid if a cancel
+	 * button got requested in the `WorkDoneProgressStart` payload.
+	 *
+	 * Clients that don't support cancellation or don't support control the button's
+	 * enablement state are allowed to ignore the setting.
+	 */
+	Boolean cancellable
+
+	/**
+	 * Optional, more detailed associated progress message. Contains
+	 * complementary information to the `title`.
+	 *
+	 * Examples: "3/25 files", "project/src/module2", "node_modules/some_dep".
+	 * If unset, the previous progress message (if any) is still valid.
+	 */
+	String message
+
+	/**
+	 * Optional progress percentage to display (value 100 is considered 100%).
+	 * If not provided infinite progress is assumed and clients are allowed
+	 * to ignore the `percentage` value in subsequent in report notifications.
+	 *
+	 * The value should be steadily rising. Clients are free to ignore values
+	 * that are not following this rule.
+	 */
+	Integer percentage;
+}
+
+/**
+ * The notification payload about progress reporting.
+ * Signaling the end of a progress reporting is done using the following payload:
+ * 
+ * Since 3.15.0
+ */
+@JsonRpcData
+class WorkDoneProgressEnd implements WorkDoneProgressNotification {
+
+	/**
+	 * Always return end
+	 */
+	override WorkDoneProgressKind getKind() {
+		WorkDoneProgressKind.end
+	}
+
+	/**
+	 * Optional, a final message indicating to for example indicate the outcome
+	 * of the operation.
+	 */
+	String message
+}
+
+/**
+ * The base protocol offers also support to report progress in a generic fashion. 
+ * This mechanism can be used to report any kind of progress including work done progress 
+ * (usually used to report progress in the user interface using a progress bar) 
+ * and partial result progress to support streaming of results. 
+ * A progress notification has the following properties:
+ * 
+ * Since 3.15.0
+ */
+@JsonRpcData
+class ProgressParams {
+
+	Either<String, Number> token
+
+	@JsonAdapter(WorkDoneProgressNotificationAdapter.Factory)
+	WorkDoneProgressNotification value;
+
+	new() {
+	}
+
+	new(@NonNull Either<String, Number> token, WorkDoneProgressNotification value) {
+		this.token = token
+		this.value = value
+	}
+
+}
+
+
+/**
+ * A parameter literal used to pass a work done progress token.
+ * 
+ * Since 3.15.0
+ */
+interface WorkDoneProgressParams {
+	/**
+	 * An optional token that a server can use to report work done progress.
+	 */
+	def Either<String, Number> getWorkDoneToken()
+	
+	def void setWorkDoneToken(Either<String, Number> token)
+	
+}
+
+/**
+ * Options to signal work done progress support in server capabilities.
+ * 
+ * Since 3.15.0
+ */
+interface WorkDoneProgressOptions {
+	
+	def Boolean getWorkDoneProgress()
+	
+	def void setWorkDoneProgress(Boolean workDoneProgress)
+	
+}
+
+/**
+ * Options to signal work done progress support in server capabilities.
+ * It is not present in protocol specification, so it's just "dry" class.
+ * 
+ * Since 3.15.0
+ */
+@JsonRpcData
+abstract class AbstractWorkDoneProgressOptions implements WorkDoneProgressOptions {
+	Boolean workDoneProgress
+} 
+
+/**
+ * Options to signal work done progress support in server capabilities and TextDocumentRegistrationOptions.
+ * It is not present in protocol specification, so it's just "dry" class.
+ * 
+ * Since 3.15.0
+ */
+@JsonRpcData
+abstract class AbstractTextDocumentRegistrationAndWorkDoneProgressOptions extends TextDocumentRegistrationOptions implements WorkDoneProgressOptions {
+	Boolean workDoneProgress
+	
+	new() {
+	}
+
+	new(List<DocumentFilter> documentSelector) {
+		super(documentSelector)
+	}
+}
+
+/**
+ * A parameter literal used to pass a partial result token.
+ * 
+ * Since 3.15.0
+ */
+interface PartialResultParams {
+	/**
+	 * An optional token that a server can use to report partial results (e.g. streaming) to
+	 * the client.
+	 */
+	def Either<String, Number> getPartialResultToken()
+	
+	def void setPartialResultToken(Either<String, Number> token)
+}
+
+/**
+ * Abstract class which implements work done progress and partial result request parameter.
+ * It is not present in protocol specification, so it's just "dry" class.
+ * 
+ * Since 3.15.0
+ */
+@JsonRpcData
+abstract class WorkDoneProgressAndPartialResultParams implements WorkDoneProgressParams, PartialResultParams {
+	
+	Either<String, Number> workDoneToken
+	
+	Either<String, Number> partialResultToken
+	
+}
+
+/**
+ * Abstract class which extends TextDocumentPosition and implements work done progress request parameter.
+ * It is not present in protocol specification, so it's just "dry" class.
+ * 
+ * Since 3.15.0
+ */
+@JsonRpcData
+abstract class TextDocumentPositionAndWorkDoneProgressParams extends TextDocumentPositionParams implements WorkDoneProgressParams {
+
+	Either<String, Number> workDoneToken
+
+	new() {
+	}
+
+	new(@NonNull TextDocumentIdentifier textDocument, @NonNull Position position) {
+		super(textDocument, position)
+	}
+
+	@Deprecated
+	new(@NonNull TextDocumentIdentifier textDocument, String uri, @NonNull Position position) {
+		super(textDocument, uri, position)
+	}
+}
+
+/**
+ * Abstract class which extends TextDocumentPosition and implements work done progress and partial result request parameter.
+ * It is not present in protocol specification, so it's just "dry" class.
+ * 
+ * Since 3.15.0
+ */
+@JsonRpcData
+abstract class TextDocumentPositionAndWorkDoneProgressAndPartialResultParams extends TextDocumentPositionAndWorkDoneProgressParams implements PartialResultParams {
+
+	Either<String, Number> partialResultToken
+	
+	new() {
+	}
+
+	new(@NonNull TextDocumentIdentifier textDocument, @NonNull Position position) {
+		super(textDocument, position)
+	}
+
+	@Deprecated
+	new(@NonNull TextDocumentIdentifier textDocument, String uri, @NonNull Position position) {
+		super(textDocument, uri, position)
+	}
+}
+
 @JsonRpcData
 class InitializeError {
 	/**
@@ -2740,7 +3076,10 @@ interface InitializeErrorCode {
  */
 @JsonRpcData
 @JsonAdapter(InitializeParamsTypeAdapter.Factory)
-class InitializeParams {
+class InitializeParams implements WorkDoneProgressParams {
+	
+	Either<String, Number> workDoneToken
+	
 	/**
 	 * The process Id of the parent process that started the server.
 	 */
@@ -2803,6 +3142,7 @@ class InitializeParams {
 	 * Since 3.6.0
 	 */
 	List<WorkspaceFolder> workspaceFolders
+
 }
 
 @JsonRpcData
@@ -3167,12 +3507,21 @@ class ReferenceContext {
 	}
 }
 
+@JsonRpcData
+class ReferenceOptions extends AbstractWorkDoneProgressOptions {
+}
+
+@JsonRpcData
+class ReferenceRegistrationOptions extends AbstractTextDocumentRegistrationAndWorkDoneProgressOptions {
+}
+
 /**
  * The references request is sent from the client to the server to resolve project-wide references for the symbol
  * denoted by the given text document position.
  */
 @JsonRpcData
-class ReferenceParams extends TextDocumentPositionParams {
+class ReferenceParams extends TextDocumentPositionAndWorkDoneProgressAndPartialResultParams {
+	
 	@NonNull
 	ReferenceContext context
 
@@ -3217,7 +3566,8 @@ class PrepareRenameResult {
  * The rename request is sent from the client to the server to do a workspace wide rename of a symbol.
  */
 @JsonRpcData
-class RenameParams extends TextDocumentPositionParams {
+class RenameParams extends TextDocumentPositionAndWorkDoneProgressParams {
+
 	/**
 	 * The new name of the symbol. If the given name is not valid the request must return a
 	 * ResponseError with an appropriate message set.
@@ -3477,7 +3827,8 @@ class ShowMessageRequestParams extends MessageParams {
  * The signature help request is sent from the client to the server to request signature information at a given cursor position.
  */
 @JsonRpcData
-class SignatureHelpParams extends TextDocumentPositionParams {
+class SignatureHelpParams extends TextDocumentPositionAndWorkDoneProgressParams {
+
 	/**
 	 * The signature help context. This is only available if the client specifies
 	 * to send this using the client capability  `textDocument.signatureHelp.contextSupport === true`
@@ -3593,7 +3944,8 @@ class SignatureHelp {
  * Signature help options.
  */
 @JsonRpcData
-class SignatureHelpOptions {
+class SignatureHelpOptions extends AbstractWorkDoneProgressOptions {
+
 	/**
 	 * The characters that trigger signature help automatically.
 	 */
@@ -3974,6 +4326,7 @@ class TextDocumentItem {
  */
 @JsonRpcData
 class TextDocumentPositionParams {
+	
 	/**
 	 * The text document.
 	 */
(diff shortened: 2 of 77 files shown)


org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/messages/progress/WorkDoneProgressNotifcation.java
```diff
@@ -0,0 +1,6 @@
+package org.eclipse.lsp4j.jsonrpc.messages.progress;
+
+public interface WorkDoneProgressNotifcation {
```
szarnekow, line 3: I didn’t do a full review but the typo caught my eye: Notif_i_cation

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -1183,6 +1186,20 @@ class TextDocumentClientCapabilities {
 	SelectionRangeCapabilities selectionRange
 }
 
+/**
+ * Window specific client capabilities.
+ */
+@JsonRpcData
+class WindowsClientCapabilities {
```
szarnekow, line 1193: Should this be WindowClientCapabilities?

org.eclipse.lsp4j/src/main/xtend-gen/org/eclipse/lsp4j/.gitignore
```diff
@@ -0,0 +1,188 @@
+/CallHierarchyCapabilities.java
```
szarnekow, line 1: This should not be part of the commit

Dufgui:
I add some option like `DeclarationOptions` but not add it in ServerCapabilities. Not sure we need it. And we cannot defined the corresponding provider with 3 types of `Either`. So what do you want ?

Dufgui:
Maybe I can add some « introduce version »? @spoenemann 
Moreover If we keep new Option class

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/messages/progress/WorkDoneProgressKind.java
```diff
@@ -0,0 +1,38 @@
...
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0,
+ * or the Eclipse Distribution License v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ * 
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ ******************************************************************************/
+package org.eclipse.lsp4j.jsonrpc.messages.progress;
+
+public enum WorkDoneProgressKind {
```
spoenemann, line 14: Please move this enum to the LSP-specific bundle (`org.eclipse.lsp4j`).

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/messages/progress/WorkDoneProgressNotification.java
```diff
@@ -0,0 +1,6 @@
+package org.eclipse.lsp4j.jsonrpc.messages.progress;
+
+public interface WorkDoneProgressNotification {
```
spoenemann, line 3: Please move this interface to `Protocol.xtend` in the LSP-specific bundle.

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/messages/progress/WorkDoneProgressNotification.java
```diff
@@ -0,0 +1,6 @@
...
+
+	WorkDoneProgressKind getKind();
```
spoenemann, line 5: We need a special Gson type adapter for this interface in order to parse incoming messages depending on the `kind` property. Maybe we can reuse this abstract class (or a variation of it): https://github.com/eclipse/sprotty-server/blob/master/org.eclipse.sprotty.server/src/main/java/org/eclipse/sprotty/server/json/PropertyBasedTypeAdapter.java

spoenemann (changes requested):
Thanks for the extensive contribution, @Dufgui!

It would be good to add tests for parsing / serializing the new message types. Parsing ProgressParams will not work without a dedicated Gson type adapter for WorkDoneProgressNotification.

Regarding DeclarationOptions, we also have the `Either3` type. But I'd be fine with postponing those changes.

Dufgui:
I add unit tests but i fail to launch them on my laptop. On eclipse I have an error:
The annotation property 'deprecationPolicy' is not declared on the annotation type 'org.eclipse.xtend.lib.annotations.Accessors'.
and some other on JsonRpcDataProcessor.xtend

And when i launch gradle on command line, it fail too.

cdietrich:
Please make sure you use a consistent Xtend version all over the place.
If you rebased your branch on master you should be able to build from command line

Dufgui:
sorry but gradle build not work because it use the deprecated osgi plugin. I don't have it in my cache so it's not working.
`Plugin with id 'osgi' not found.`

And on eclipse side i have 2.22. I don't find any update site with 2.19.
So i cannot fix it. any idea ?

cdietrich:
how do you run gradle. if you use gradlew it should work fine

cdietrich:
you can find all versions of xtend here: 
https://download.eclipse.org/modeling/tmf/xtext/updates/composite/releases/
or you download this one.

https://www.eclipse.org/downloads/packages/release/2019-09/r/eclipse-ide-java-and-dsl-developers

Dufgui:
that's it. I found many bugs thanks. Rest two unit tests fail. I  still investigate

Dufgui:
all tests succeed. Ready for new review cycle. And maybe merge ;o)

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/WorkDoneProgressNotificationAdapter.java
```diff
@@ -0,0 +1,175 @@
...
+import com.google.gson.TypeAdapter;
+import com.google.gson.TypeAdapterFactory;
+import com.google.gson.reflect.TypeToken;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
+import com.google.gson.stream.JsonWriter;
+import com.google.gson.stream.MalformedJsonException;
+
+/**
+ * A type adapter that reads every input into a tree of {@link JsonElement}s.
```
spoenemann, line 35: The comment does not match what this type adapter does.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/LanguageClient.java
```diff
@@ -141,4 +143,23 @@
 	default void semanticHighlighting(SemanticHighlightingParams params) {
 		throw new UnsupportedOperationException();
 	}
+
+	/**
+	 * This request is sent from the server to the client to ask the client to create a work done progress.
+	 */
+	@JsonNotification("window/workDoneProgress/create")
+	default void createProgress(WorkDoneProgressCreateParams params) {
```
spoenemann, line 151: This should be a `@JsonRequest`, and the return type should be `CompletableFuture<Void>`.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/LanguageServer.java
```diff
@@ -87,4 +88,12 @@ default void initialized() {
 	@JsonDelegate
 	WorkspaceService getWorkspaceService();
 	
+	/**
+	 * This request is sent from the server to the client to ask the client to create a work done progress.
```
spoenemann, line 92:
```suggestion
	 * This notification is sent from the client to the server to cancel a progress initiated on the server side.
```

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -1321,7 +1348,12 @@ class CodeActionContext {
  * These commands are typically code fixes to either fix problems or to beautify/refactor code.
  */
 @JsonRpcData
-class CodeActionParams {
+class CodeActionParams implements WorkDoneProgressParams, PartialResultParams {
+	
+	Either<String, Number> workDoneToken
```
spoenemann, line 1353: Given that this type is referenced very often, you could try to define
```xtend
class ProgressToken extends Either<String, Number> {
	override toString() {
		if (isLeft)
			left
		else if (isRight)
			right.toString
	}
}
```
and replace `Either<String, Number>` with this new class (i.e. redefine the type instead of wrapping it as in the previous version). I think parsing/serializing should still work the same with the EitherTypeAdapter.

spoenemann: You would also have to create a protected constructor and static creation methods as in the superclass (but I'd call them both `create` instead of `forLeft` and `forRight`).

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/LanguageClient.java
spoenemann: It's specified as a request here:
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#window_workDoneProgress_create

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/LanguageClient.java
```diff
@@ -141,4 +143,23 @@
...
+	}
+	
+	/**
+	 * The base protocol offers also support to report progress in a generic fashion. 
+	 * This mechanism can be used to report any kind of progress including work done progress 
+	 * (usually used to report progress in the user interface using a progress bar) 
+	 * and partial result progress to support streaming of results.
+	 */
+	@JsonRequest("$/progress")
+	default CompletableFuture<Void> notifyProgress(ProgressParams params) {
```
spoenemann, line 162: This should remain a `@JsonNotification`, with `void` return type:
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#progress

My comment about `@JsonRequest` was related only to `window/workDoneProgress/create`.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/ProgressTokenAdapter.java
```diff
@@ -0,0 +1,87 @@
...
+import com.google.gson.TypeAdapterFactory;
+import com.google.gson.reflect.TypeToken;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
+import com.google.gson.stream.JsonWriter;
+
+/**
+ * A type adapter for class which implements WorkDoneProgressNotification interface.
+ */
+public class ProgressTokenAdapter extends TypeAdapter<ProgressToken> {
```
spoenemann, line 30: Is this adapter really necessary? We already have a generic type adapter for `Either`, which should work for `ProgressToken` without further configuration (Gson considers the superclass of `ProgressToken` while looking for a type adapter). Please try whether parsing / serializing works without this adapter and the corresponding `@JsonAdapter` annotations.

spoenemann (commented):
Almost there :-D

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/ProgressTokenAdapter.java
Dufgui: If you want it, it's possible but this way i must put ProgressToken class in json project because the EitherTypeAdapter is in json project. See my last separated commit 
https://github.com/eclipse/lsp4j/pull/442/commits/8f634798a5308be1500f1b0da1328def5caa7ad5

Or maybe you have another idea ?

spoenemann: I see. I think the following should work:

 * Add the following code at the start at `createLeft` and `createRight` of EitherTypeAdapter (pass `null, obj` instead of `obj, null` in `createRight`):
```java
try {
	Constructor<Either<L, R>> constructor = (Constructor<Either<L, R>>) typeToken.getRawType().getConstructor(Object.class, Object.class);
	return constructor.newInstance(obj, null);
} catch (NoSuchMethodException | InstantiationException | IllegalAccessException | InvocationTargetException exc) {
	// Proceed with the code below
}
```
 * Make the constructor of `ProgressToken` public.

This way we avoid the dependency of EitherTypeAdapter to ProgressToken and you can leave ProgressToken in the LSP-specific bundle.

Dufgui: et voila !!!

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -1321,7 +1349,14 @@ class CodeActionContext {
  * These commands are typically code fixes to either fix problems or to beautify/refactor code.
  */
 @JsonRpcData
-class CodeActionParams {
+class CodeActionParams implements WorkDoneProgressParams, PartialResultParams {
+	
+	@JsonAdapter(EitherTypeAdapter.Factory)
```
spoenemann, line 1354: EitherTypeAdapter is registered globally in Gson. You should be able to remove these annotations and it should work the same.

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/EitherTypeAdapter.java
```diff
@@ -201,16 +203,30 @@ public void write(JsonWriter out, Either<L, R> value) throws IOException {
 	protected Either<L, R> createLeft(L obj) throws IOException {
 		if (Either3.class.isAssignableFrom(typeToken.getRawType()))
 			return (Either<L, R>) Either3.forLeft3(obj);
-		else
-			return Either.forLeft(obj);
+		else if ("org.eclipse.lsp4j.ProgressToken".equals(typeToken.getRawType().getName())) {
```
szarnekow, line 206: This looks a little spooky to me. If you follow the original suggestion by @spoenemann and remove the type ProgressToken, this hidden inverse dependency would become obsolete. Why do we need the ProgressToken type at all?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -2390,11 +2447,28 @@ class ResolveTypeHierarchyItemParams {
...
+}
+
 /**
  * The document symbol request is sent from the client to the server to list all symbols found in a given text document.
  */
 @JsonRpcData
-class DocumentSymbolParams {
+class DocumentSymbolParams implements WorkDoneProgressParams, PartialResultParams {
+	
+	@JsonAdapter(EitherTypeAdapter.Factory)
```
szarnekow, line 2466: My gut feeling is that we should introduce a base class for all `*Params` that implement `WorkDoneProgressParams, PartialResultParams` and move the fields into that base class. I see that the spec does this differently, but it's only defining the interface and it could be worthwhile to reduce the duplication on the impl side, too.
This would also allow for generic progress implementations without the need to define a signature like
`<T extends WorkDoneProgressParams & PartialResultParams> myMethod(T paramsWithProgress)`

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -2703,6 +2777,175 @@ class MarkedString {
 	}
 }
 
+interface WorkDoneProgressNotification {
```
szarnekow, line 2780: Please add a comment why this was introduced.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -2703,6 +2777,175 @@ class MarkedString {
...
+	}
+
+	/**
+	 * Optional, a final message indicating to for example indicate the outcome
+	 * of the operation.
+	 */
+	String message
+}
+
+class ProgressToken extends Either<String, Number> {
```
szarnekow, line 2878: As stated above, I don't think this carries its own weight. Also extending `Either` does not seem to be a good choice. One would usually also not extend `ArrayList` or `HashMap` and I picture `Either` to be a data type that plays in the same ballpark as the types from java.util.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -2703,6 +2777,175 @@ class MarkedString {
...
+	/**
+	 * An optional token that a server can use to report work done progress.
+	 */
+	def ProgressToken getWorkDoneToken()
+	
+	def void setWorkDoneToken(ProgressToken token)
+	
+}
+
+interface WorkDoneProgressOptions {
```
szarnekow, line 2931: Please add the (sparse) documentation from the spec here, too.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -3167,12 +3415,26 @@ class ReferenceContext {
...
+	Boolean workDoneProgress
+}
+
 /**
  * The references request is sent from the client to the server to resolve project-wide references for the symbol
  * denoted by the given text document position.
  */
 @JsonRpcData
-class ReferenceParams extends TextDocumentPositionParams {
+class ReferenceParams extends TextDocumentPositionParams implements WorkDoneProgressParams, PartialResultParams {
```
szarnekow, line 3433: Is the `implements WorkDoneProgressParams` redundant?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -4406,11 +4680,35 @@ class WorkspaceEdit {
 	}
 }
 
+
+/**
+ * The options of a Workspace Symbol Request.
+ */
+@JsonRpcData
+class WorkspaceSymbolOptions implements WorkDoneProgressOptions {
+	Boolean workDoneProgress
```
szarnekow, line 4689: Similar remark as above: Should we have a base type with the field declaration and the other `*Options` should inherit the field from that one?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -4743,25 +5046,20 @@ class ExecuteCommandParams {
 		this.command = Preconditions.checkNotNull(command, 'command')
 		this.arguments = arguments
 	}
+	
+	new(@NonNull String command, List<Object> arguments, ProgressToken workDoneToken) {
+		this.workDoneToken = workDoneToken
```
szarnekow, line 5051: Nit: For consistency please assign the fields in the same order as the parameter lists.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -2703,6 +2777,175 @@ class MarkedString {
 	}
 }
 
+interface WorkDoneProgressNotification {
+	def WorkDoneProgressKind getKind()
+}
+ 
+
+@JsonRpcData
+class WorkDoneProgressBegin implements WorkDoneProgressNotification {
```
szarnekow, line 2786: Please add `Since` information as JavaDoc similar to how it was done for other types.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/WorkDoneProgressKind.java
```diff
@@ -0,0 +1,43 @@
...
+ * http://www.eclipse.org/legal/epl-2.0,
+ * or the Eclipse Distribution License v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ * 
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ ******************************************************************************/
+package org.eclipse.lsp4j;
+
+public enum WorkDoneProgressKind {
+	BEGIN("begin"),
```
szarnekow, line 15: Maybe it's worthwhile to use lowercase enum literals here which would render basically the entire logic in this enum declaration obsolete.
´public enum WorkDoneProgressKind { begin, report, end; }`

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/WorkDoneProgressKind.java
```diff
@@ -0,0 +1,43 @@
...
+	
+	WorkDoneProgressKind(String value) {
+		this.value = value;
+	}
+	
+	public String getValue() {
+		return value;
+	}
+	
+	public static WorkDoneProgressKind forValue(String value) {
```
szarnekow, line 29: `Enum.valueOf(..)` and `forValue` may become a source of confusion.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/WorkDoneProgressKind.java
```diff
@@ -0,0 +1,43 @@
...
+		for (WorkDoneProgressKind workDoneProgressKind : allValues) {
+			if(workDoneProgressKind.getValue().equals(value)) {
+				return workDoneProgressKind;
+			}
+		}
+		throw new IllegalArgumentException("Illegal enum value: " + value);
+	}
+
+	@Override
+	public String toString() {
```
szarnekow, line 40: Not sure about this. The doc allows to override toString() but it kind-of suggests that it should not be necessary:
```
Returns the name of this enum constant, as contained in the declaration. This method may be overridden, though it typically isn't necessary or desirable. An enum type should override this method when a more "programmer-friendly" string form exists.
```

org.eclipse.lsp4j/src/test/java/org/eclipse/lsp4j/test/services/JsonParseTest.xtend
```diff
@@ -1417,4 +1422,76 @@ class JsonParseTest {
...
+	
+	@Test
+	def void testProgressCreate() {
+		'''
+			{
+				"jsonrpc": "2.0",
+				"id": 1,
+				"method": "window/workDoneProgress/create",
+				"params": {
+					"token": "progress-token"
```
szarnekow, line 1436: Please add tests with tokens that are numbers.

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/EitherTypeAdapter.java
spoenemann: > Why do we need the ProgressToken type at all?

I proposed that because the type is used very often in the protocol. It's just a type alias (which does not exist in Java) that makes the protocol implementation more readable.

We can remove the class name check and always try to invoke the public constructor if it exists.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
spoenemann: I see this differently. Supporting the creation of Either subclasses via a public constructor is a small change. Using the type `ProgressToken` in a field makes the protocol much better readable compared to `Either<String, Number>`.

However, if we can reduce the references to this type to only a few via an abstract superclass as you proposed above, the situation would be different.

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/EitherTypeAdapter.java
Dufgui: In my opinion i prefere something more simple:
- no ProgressToken and replace the Either<String, Number>, like i do before
- If you absolutely want this class, i propose to add an adapter, this way the code is more simple and less hugly.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
Dufgui: ok for me, @spoenemann ?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
Dufgui: Sorry i don't catch your last sentence.

> However, if we can reduce the references to this type to only a few via an abstract superclass as you proposed above, the situation would be different.

What is the proposition ? Create an abstract class common between Either and ProgressToken ?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/WorkDoneProgressKind.java
Dufgui: ok for me, definitly not used to use lower on enum, but why not

org.eclipse.lsp4j/src/test/java/org/eclipse/lsp4j/test/services/JsonParseTest.xtend
Dufgui: ok

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/EitherTypeAdapter.java
spoenemann: Ok then revert back to `Either<String, Number>` as before.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
spoenemann: :+1:

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
spoenemann: I meant this proposal: https://github.com/eclipse/lsp4j/pull/442#discussion_r476240366

Dufgui:
hello. I re align the spec and code (some params must not have progress on it)
and i fix all your return from code review. Tests are green too ;o)

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -1321,7 +1348,8 @@ class CodeActionContext {
  * These commands are typically code fixes to either fix problems or to beautify/refactor code.
  */
 @JsonRpcData
-class CodeActionParams {
+class CodeActionParams extends WorkDoneProgressAndPartialParams {
```
szarnekow, line 1351: Nit: Should the type be called `WorkDoneProgressAndPartialResultParams` instead of `WorkDoneProgressAndPartialParams`?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -2313,7 +2349,10 @@ class DocumentOnTypeFormattingParams extends DocumentFormattingParams {
  * The document range formatting request is sent from the client to the server to format a given range in a document.
  */
 @JsonRpcData
-class DocumentRangeFormattingParams extends DocumentFormattingParams {
+class DocumentRangeFormattingParams extends DocumentFormattingParams implements WorkDoneProgressParams {
```
szarnekow, line 2352: Redundant super type? The spec say `interface DocumentFormattingParams extends WorkDoneProgressParams {..}` so DocumentRangeFormattingParams should not need the super type again.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -2390,11 +2429,22 @@ class ResolveTypeHierarchyItemParams {
 
 }
 
+@JsonRpcData
+class DocumentSymbolOptions implements WorkDoneProgressOptions {
```
szarnekow, line 2433: Many of the `*Options` implement `WorkDoneProgressOptions` and define the `Boolean workDoneProgress`. Would a super class make sense?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -2703,6 +2753,210 @@ class MarkedString {
 	}
 }
 
+/**
+ *  the $/progress notification payload interface.
```
szarnekow, line 2757: Nit: Please start the sentence with uppercase `The`

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -2703,6 +2753,210 @@ class MarkedString {
...
+ *
+ * Since 3.15.0
+ */
+interface WorkDoneProgressNotification {
+	def WorkDoneProgressKind getKind()
+}
+ 
+/**
+ *  the $/progress notification payload to start progress reporting.
```
szarnekow, line 2766: Nit: Please start the sentence with uppercase `The`

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -2703,6 +2753,210 @@ class MarkedString {
...
+ *
+ * Since 3.15.0
+ */
+@JsonRpcData
+class WorkDoneProgressBegin implements WorkDoneProgressNotification {
+
+	override WorkDoneProgressKind getKind() {
+		WorkDoneProgressKind.begin
```
szarnekow, line 2774: I like the fact that the enum literals are exactly spelled as in the spec.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -2703,6 +2753,210 @@ class MarkedString {
...
+ 
+/**
+ *  the $/progress notification payload to start progress reporting.
+ *
+ * Since 3.15.0
+ */
+@JsonRpcData
+class WorkDoneProgressBegin implements WorkDoneProgressNotification {
+
+	override WorkDoneProgressKind getKind() {
```
szarnekow, line 2773: Please add a comment that this is always returning of the kind `begin`

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -2703,6 +2753,210 @@ class MarkedString {
...
+	 * to ignore the `percentage` value in subsequent in report notifications.
+	 *
+	 * The value should be steadily rising. Clients are free to ignore values
+	 * that are not following this rule.
+	 */
+	Integer percentage;
+}
+
+/**
+ *  The notification payload about progress reporting.
```
szarnekow, line 2855: Nit: Double space at the beginning of the comment.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -2703,6 +2753,210 @@ class MarkedString {
...
+ * 
+ * Since 3.15.0
+ */
+@JsonRpcData
+class ProgressParams {
+
+	Either<String, Number> token
+
+	@JsonAdapter(WorkDoneProgressNotificationAdapter.Factory)
+	WorkDoneProgressNotification value;
```
szarnekow, line 2889: Not sure if this was already discussed:
```
interface ProgressParams<T> {
	/**
	 * The progress token provided by the client or server.
	 */
	token: ProgressToken;

	/**
	 * The progress data.
	 */
	value: T;
}
```

What is the purpose of the type parameter `T` in the spec and are we missing an opportunity / a valid use case by typing this to `WorkDoneProgressNotification`?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -2703,6 +2753,210 @@ class MarkedString {
...
+	 * An optional token that a server can use to report work done progress.
+	 */
+	def Either<String, Number> getWorkDoneToken()
+	
+	def void setWorkDoneToken(Either<String, Number> token)
+	
+}
+
+/**
+ *  Options to signal work done progress support in server capabilities.
```
szarnekow, line 2918: Nit: Double space at the beginning of the comment

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -2703,6 +2753,210 @@ class MarkedString {
...
+	 * An optional token that a server can use to report partial results (e.g. streaming) to
+	 * the client.
+	 */
+	def Either<String, Number> getPartialResultToken()
+	
+	def void setPartialResultToken(Either<String, Number> token)
+}
+
+/**
+ *  Abstract class which implements work done progress and partial result request parameter.
```
szarnekow, line 2946: Looks like some of the comment formatting is a little off. Not super important. Autoformat does create too much noise I suppose?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
szarnekow: No, it's not.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -4008,8 +4289,14 @@ class TextDocumentPositionParams {
 	}
 }
 
+
 @JsonRpcData
-class CompletionParams extends TextDocumentPositionParams {
+class CompletionParams extends TextDocumentPositionParams implements WorkDoneProgressParams, PartialResultParams {
```
szarnekow, line 4294: Would a `abstract class TextDocumentPositionParamsWithProgress extends TextDocumentPositionParams implements .. { .. }` make sense?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
szarnekow: This is basically the same remark as above for the `*Options`

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/WorkDoneProgressKind.java
```diff
@@ -0,0 +1,17 @@
+/******************************************************************************
+ * Copyright (c) 2016 TypeFox and others.
```
szarnekow, line 2: Wrong copyright year.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/WorkDoneProgressNotificationAdapter.java
```diff
@@ -0,0 +1,175 @@
+/******************************************************************************
+ * Copyright (c) 2018 TypeFox and others.
```
szarnekow, line 2: Please use the correct year / copyright information.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/WorkDoneProgressNotificationAdapter.java
```diff
@@ -0,0 +1,175 @@
...
+			report.setMessage(message);
+			report.setCancellable(cancellable);
+			report.setPercentage(percentage);
+			return report;
+		case "end":
+			WorkDoneProgressEnd end = new WorkDoneProgressEnd();
+			end.setMessage(message);
+			return end;
+		default:
+			throw new MalformedJsonException("Kind of progress notification is unknown :"+kind);
```
szarnekow, line 132: The colon at the end of the error message should be before the space and not after the space.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/WorkDoneProgressNotificationAdapter.java
```diff
@@ -0,0 +1,175 @@
...
+		}
+	}
+
+	@Override
+	public void write(JsonWriter out, WorkDoneProgressNotification notif) throws IOException {
```
szarnekow, line 137: `notif`?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/WorkDoneProgressNotificationAdapter.java
```diff
@@ -0,0 +1,175 @@
...
+			out.value(report.getPercentage());
+			break;
+		case end:
+			WorkDoneProgressEnd end = (WorkDoneProgressEnd) notif;
+			out.name("message");
+			out.value(end.getMessage());
+			break;
+
+		default:
+			break;
```
szarnekow, line 170: Throw exception in case there will be new notification kinds I'd like to throw it into our face if the adapter is not updated accordingly.

szarnekow (commented):
I like the direction this is heading to and the progress. We are certainly getting there. 
While reviewing the protocol changes, I tried to match the type definitions against the spec. For the formatting params, I think something slipped through the cracks.

The artificial base type makes for `WorkDoneProgressAndPartialResultParams` pans out nicely - at least that's my impression. We can take this a step further by using a similar articial "dry" type for the `TextDocumentPositionParams` with `WorkDoneProgressParams` and `TextDocumentPositionParams` with `WorkDoneProgressParams` and `PartialResultParams`. I think it would make sense to avoid more duplications.
What do you think about the proposed change to the `Options` - meaning to have a `abstract clss AbstractWorkDoneProgressOptions` that defines the `Boolean` field?

Some of the other feedback is merely stylistic and if it bothers you to get feedback on whitespaces in JavaDoc, feel free to tell me.

szarnekow:
@Dufgui please not that your commit needs a proper sign-off footer.

Dufgui:
for the sign off i think do it at the end when i squash my commit. ;o)

for the abstract classes, i do it in my first merge request.... so yes ok absolutely. ;o)

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
Dufgui: It's the previous code of the protocol. Yes your are right.
We can change it here, or a create new merge request dedicated., up to you

Dufgui:
any news ?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
spoenemann: It's strange that they didn't constrain `T` to any supertype. But Section [Work Done Progress](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workDoneProgress) states

> The value payload of a work done progress notification can be of three different forms.

These "different forms" are exactly the three implementations of the WorkDoneProgressNotification interface. I think this is a sufficient approximation of what they are trying to express in the spec.

spoenemann:
@Dufgui thank you for your endurance :-D You chose a complex topic, but keep up the good work! :+1:

There are a few unresolved comments in Protocol.xtend, could you address them?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
Dufgui: And in fact generics will add many <?> in the code. I am not sur it's a good idea.

Dufgui:
hope i don't miss any comment. I don"t see any review comment not take account.

spoenemann (approved):
Great work @Dufgui!

PR #443, opened 2020-08-21
fix #435 align to spec

### LINKED ISSUE #435: TextDocumentService's declaration and definition do not support raw locations in the response
In the TextDocumentService, both the goto definition request and goto declaration request are written to return either a list of Locations or a list of LocationLinks. 

However, in the [LSP specification](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_definition), a valid return value is a raw Location object. If the language server returns a raw Location object, however, it causes the response parser to choke as it is not expected.

Dufgui:
Fix #435


Dufgui:
ready for review

Dufgui:
any reviewer ?

spoenemann (commented):
Sorry for the late review.

I don't think we need to go this complicated way, especially because it would break the API and force all consumers of LSP4J to adapt.

The four affected JSON-RPC methods are already using LocationLinkListAdapter to parse their result values. That adapter uses a combination of ListChecker and PropertyChecker to determine whether the result is a list of Location or a list of LocationLink. The type adapter is then an EitherTypeAdapter with CollectionTypeAdapter for both alternatives. In theory, this setup should be able to parse a single value (not wrapped in an array). If that doesn't work, we should debug to find out why.

spoenemann (commented):
Great catch!

Could you move the new test into a separate method? Then it's good to merge.

spoenemann (approved):
Thanks!

PR #449, opened 2020-10-07
fix #447 add progress on call hierarchy

Dufgui:

PR #451, opened 2020-10-08
Fix Preconditions

Dufgui:
Signed-off-by: Guillaume Dufour <guillaume.duff@gmail.com>

PR #452, opened 2020-10-11
Params updates

KamasamaK:
Went through all of the Params and made some updates.

- Updated `DocumentFormattingParams` and `DocumentRangeFormattingParams` to support progress. `DocumentOnTypeFormattingParams` now extends `TextDocumentPositionParams` instead of `DocumentFormattingParams` because it does not support progress and better matches the spec. It's possible this may have compatibility issues, but seems unlikely. `DocumentRangeFormattingParams` also no longer extends `DocumentFormattingParams` to better match the spec and prevent future compatibility issues.
- Added `@Beta` to  some `TypeHierarchy` classes that were missing it
- Removed deprecated constructors for new classes just added in #442 since they shouldn't be used
- Added and fixed some comments
- Made spacing more consistent
- Skipped `CallHierarchy` Params to prevent conflict with #449

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -1473,7 +1473,6 @@ class CodeActionContext {
  */
 @JsonRpcData
 class CodeActionParams extends WorkDoneProgressAndPartialResultParams {
-
 	/**
 	 * The document in which the command was invoked.
 	 */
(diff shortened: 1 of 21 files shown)


KamasamaK:
#451 fixes the failing tests.

KamasamaK:
@eclipse-lsp4j-bot run tests

Dufgui:
https://ci.eclipse.org/lsp4j/job/lsp4j-github-pullrequests/129/ test are ok.
Seems ok. 
I propose only to add `@since ` on TypeHierarchy. But i don't find them on specification 3.16. Where it comme from ?

KamasamaK:
@Dufgui `TypeHierarchy` is nonstandard
https://github.com/eclipse/lsp4j/blob/8b357e6045c660034e14ee2473bcbd206d72141f/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend#L993-L995

Dufgui:
As i can see it's since 3.15.0 but not yet integrated, your are true.

Dufgui:
So, ok for me

PR #454, opened 2020-10-11
Fix list of token formats on semantic tokens capabilities

### LINKED ISSUE #453: semanticTokens client capability sends unexpected data
Hello.

I've tried to run simple example with semanticToken responses, but I can't establish connection between VSCode Language Client version `7.0.0-next.9` and LSP4J, build from `5ff540af5b0f1518dc9f61d0b9673c3f513c8724` commit.

I've got next error message in VSCode logs comes from my LS:
```
{"jsonrpc":"2.0","id":0,"error":{"code":-32700,"message":"Message could not be parsed.","data":{"message":"java.lang.IllegalStateException: Expected BEGIN_OBJECT but was STRING at line 1 column 3574 path $.params.capabilities.textDocument.semanticTokens.formats[0]"}}}
```

I can't trace full outbound request at the moment, but I guess, that VSCode sends `formats` field from the spec

```typescript
	/**
	 * The formats the clients supports.
	 */
	formats: TokenFormat[];
```

as array of `string`, not an array of `object`.

`TokenFormat` itself described as alias over `string`:

```typescript
export namespace TokenFormat {
	export const Relative: 'relative' = 'relative';
}

export type TokenFormat = 'relative';
```

And it looks like, that LSP4J wants an object, and can't deserialize string to `TokenFormat` from https://github.com/eclipse/lsp4j/blob/master/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/TokenFormat.java

ericdallo:
FIxes #453

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -1127,16 +1127,18 @@ class SemanticTokensCapabilities extends DynamicRegistrationCapabilities {
 	List<String> tokenModifiers;
 
 	/**
-	 * The formats the clients supports.
+	 * The tokens the client supports.
+	 *
+	 * See {@link TokenFormat} for allowed values.
 	 */
 	@NonNull
-	List<TokenFormat> formats;
+	List<String> formats;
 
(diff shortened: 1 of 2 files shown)


ericdallo:
FYI @nixel2007

spoenemann (approved):
Thanks!

Issue #457, opened 2020-10-20
Removing Semantic Highlighting


KamasamaK:
> Given [this](https://github.com/Microsoft/vscode-languageserver-node/pull/367), should we remove or deprecate semantic highlights as semantic tokens replace it?

_Originally posted by @ericdallo in https://github.com/eclipse/lsp4j/pull/446#issuecomment-683374512_

This question was never addressed, so I'm creating a separate issue for it. Now that Semantic Tokens are officially in the LSP 3.16 proposal and implemented in LSP4J, the Semantic Highlighting operations should be removed or deprecated. What is the view on beta/experimental features? Should they just be removed or should they be deprecated first even though they were never finalized?

spoenemann:
The `@Beta` annotation says something like "use at your own risk, this might change or be removed". Nonetheless it would make sense to add `@Deprecated` to the affected parts of the API before our 0.10.0 release and remove them after the release.

KamasamaK:
I've added `@Deprecated` to the Semantic Highlighting API. I will remove that API some time after 0.10.0 release.

KamasamaK:
@kittaakos As the one who created the semantic highlighting API, how do you feel about the timeline of its removal?

kittaakos:
@KamasamaK, thank you for keeping an eye on this. :+1: Please, feel free to remove the unused APIs. I do not have any obejctions.

PR #461, opened 2020-10-27
Update CHANGELOG.md

spoenemann:
This is a preparation for the [upcoming release 0.10.0](https://projects.eclipse.org/projects/technology.lsp4j/releases/0.10.0). 

We should track all breaking changes. @KamasamaK @Dufgui @ericdallo could you help with this? Are you aware of any breaking changes that came with your contributions?

CC @jonahgraham

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### v0.10.0 (Nov. 2020)
 
+ * Implemented LSP version 3.16.0 (specification is not finalized yet)
  * Implemented DAP version 1.37.0's CancelRequest
  * Implemented DAP version 1.38.0
  * Implemented DAP version 1.39.0
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ The snapshots are deployed during the build to:
 
 ### Supported LSP Versions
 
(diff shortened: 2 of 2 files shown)


ericdallo:
About semanticTokens side, AFAIK this would be the only breaking change: https://github.com/eclipse/lsp4j/issues/457 but it was not implemented yet (and maybe it should be for the next release?)

jonahgraham:
Thanks @spoenemann for raising this. 

The readme also needs an update: https://github.com/eclipse/lsp4j/blob/0faac0cc5d1dc51f45b9c7540ad9b94eef38ddca/README.md#supported-lsp-versions

KamasamaK:
The only potentially breaking change with my contributions is that `DocumentOnTypeFormattingParams` now extends `TextDocumentPositionParams` instead of `DocumentFormattingParams`. Also, `DocumentRangeFormattingParams` no longer extends `DocumentFormattingParams`. I doubt that would actually result in breaking anything, but I suppose it's possible.

Dufgui:
I think I don’t have breaking change.
I only append fields or parent class. I don’t remove anything.

spoenemann:
Thanks!

PR #464, opened 2020-11-03
[releng] Remove Jenkinsfile that is not used

jonahgraham:
Current Jenkinsfiles are:

releng/release-eclipse.Jenkinsfile
releng/build.Jenkinsfile


jonahgraham:
@spoenemann I assume you don't use this anymore. Just checking before I submit this change.

spoenemann (approved):
I think this has been used by

http://services.typefox.io/open-source/jenkins/job/lsp4j/

but we don't need that anymore.

Issue #199, opened 2018-06-27
On first opening or cleaning projects non-modified files are showing up with changes


jonahgraham:
Doing a simple `./gradlew build eclipse` causes numerous files to have diffs. I presume this is due to differences between versions used to originally create files and current versions.

Similarly, doing a clean in Eclipse causes some java files to be generated with different content.

I would like to submit *my* version of the files to the repo so that I don't see these diffs, but I expect that means that someone else will have the problem.

Any suggestions on best way to deal with this?

pisv:
This almost made me cry as a first time contributor.

Before discovering this description of the issue, I thought there was something wrong with my Eclipse installation and relentlessly tried to find a proper combination of Buildship and Xtend IDE going back as far as Oxygen without success.

One way to solve it might be to pinpoint the versions of Buildship and Xtend IDE that everyone should use and regenerate everything using these versions. The information about the used versions could then be stated in the contribution guide. Also, the required tool versions could be described in a `.p2f` file.

jonahgraham:
I raised PR#492 to better document what is happening with `xtend-gen`. 

> One way to solve it might be to pinpoint the versions of Buildship and Xtend IDE that everyone should use and regenerate everything using these versions. 

The key thing is that `./gradlew build` is the "correct" one to use. I am not sure what combination in the IDE that is. It is annoying that xtend-gen directories differ.

It may be better to have the discussion about why xtend-gen directories are checked in. @spoenemann Why are the xtend-gen directories checked in? I saw that you a while ago stopped doing that for test directories. I assume there is some value in having them in source control, and if so I would like to document that as part of PR#492.

spoenemann:
I think it should be fine to remove the generated code from source control. I don't remember why we decided to include it at some point.

pisv:
> I raised PR#492 to better document what is happening with `xtend-gen`.
> The key thing is that `./gradlew build` is the "correct" one to use.

Oh, I see now. I missed that part, and instead of using the command line, tried to follow the steps described in the Eclipse section of the contributing guide, with mixed results. Perhaps it would not be an issue for someone who knows the gradle/Buildship/Xtend stack a bit more than I do. Thanks a lot for the detailed description in #492, I really appreciate it.

Not keeping the generated code under source control seems reasonable indeed.

jonahgraham:
> I think it should be fine to remove the generated code from source control. I don't remember why we decided to include it at some point.

[...]

PR #506, opened 2021-02-11
Bug #199: Remove .project files from git

jonahgraham:
These files are created when importing into Eclipse IDE so
should not be needed in source control.


jonahgraham:
The flow looks better without the .project files in source control. 

There is one minor issue with removing them, while handling the transition to the removal, git checkout of commits with and without the .project files can be a pain. For example, devs doing a git pull once this PR is merged will have their projects in the IDE go "wrong", so some extra care is needed. 

Long term this is a better solution - but I will leave this PR open for a few days in case someone has an issue with this change.

jonahgraham:
> Long term this is a better solution - but I will leave this PR open for a few days in case someone has an issue with this change.

No comment for 5 days. I will merge it now.

Issue #510, opened 2021-02-17
testVersatility test unstable because of how PipedInputStream and PipedOutputStream is used


jonahgraham:
While investigation unstable tests in #508 @cdietrich identified testVersatility is unstable depending on the number of processors.

This is due to a misuse of PipedInputStream and PipedOutputStream in the test - caused by a bug in the Java library. Full details are in https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-4028322, but the summary is:

> The PipedInputStream class does not always work properly when more than one thread writes to the corresponding PipedOutputStream.

In the case of LSP4J there are multiple threads writing to the PipedOutputStream.

jonahgraham:
And why is this running on a different thread? Because testVersatility is the only test that tests exceptions. But execeptionally does not run on the thread pool's thread, but rather on the thread the exception was created. The thread that is created in this case is by a call to `CompletableFutures.computeAsync` in `MyClient`:

https://github.com/eclipse/lsp4j/blob/ad99a3d77139351dbc00c2dacd388fe749652309/org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/IntegrationTest.java#L357-L361

Of course it would be nice if `exceptionally` was called on the same thread as the non-exceptional processing would have happened, but that is another JDK issue covering it: https://bugs.openjdk.java.net/browse/JDK-8046356. The workaround is:

> To guarantee that the exceptionally task is executed on a thread within the executor thread pool then it is necessary to use whenCompleteAsync or handleAsync and passing in the executor.

That means changing https://github.com/eclipse/lsp4j/blob/ad99a3d77139351dbc00c2dacd388fe749652309/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/RemoteEndpoint.java#L282

However, once you go down that path, you see that a few lines up the `thenAccept` also may run on an arbitrary thread as the executor is not passed to it either. https://github.com/eclipse/lsp4j/blob/ad99a3d77139351dbc00c2dacd388fe749652309/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/RemoteEndpoint.java#L279

[...]

PR #511, opened 2021-02-17
Fix timeout and threading issues in tests

jonahgraham:


jonahgraham:
Submitted without all checks passing because the checks got past the problem #510 was trying to address.

Issue #508, opened 2021-02-16
Migrate LSP4J to new Eclipse Foundation Build Infra


jonahgraham:
The EF is in the process of moving LSP4J to new Build Infra. See https://bugs.eclipse.org/bugs/show_bug.cgi?id=570632

jonahgraham:
The migration step 1 is done, we have a new build machine https://ci-staging.eclipse.org/lsp4j/ - before moving that to the https://ci.eclipse.org/lsp4j/ URL I need to make sure it looks like it works properly.

I have disabled builds on the old CI to prevent double builds. I have kicked off a master build and a PR build (for #507) so that I can make sure the new CI works.

jonahgraham:
All builds on ci-staging failing at this point:
```
Task :org.eclipse.lsp4j.jsonrpc:test
```

cdietrich:
i assume we need a specific node like
https://github.com/eclipse/Xpect/blob/0e5186a5fd96d4c82536fcff9acfa6ffada9fff8/Jenkinsfile#L13

jonahgraham:
> i assume we need a specific node like
> https://github.com/eclipse/Xpect/blob/0e5186a5fd96d4c82536fcff9acfa6ffada9fff8/Jenkinsfile#L13

I hope not - I am using the standard as documented [here](https://wiki.eclipse.org/Jenkins#How_do_I_run_a_Java.2FMaven_build_on_the_new_infra.3F) which does not request a specific node. I may need to add specific tools though.

cdietrich:
we had to change the xpect build to the centos recently

jonahgraham:
I'll try it on there instead of `basic` to see if that "just works". If it does work, it means I will have to convert the PR job over to a Jenkinsfile as it implicitly runs on `basic` now as a freestyle job,

jonahgraham:
I saw you are doing some tests - thank you. Please let me know what you discover. I have to sign off now until later today (or tomorrow).

cdietrich:
maybe we also need some explicit memory settings for the build.
the old one used JAVA_TOOL_OPTIONS with Xmx2G
am not sure if this need to be configured on Jenkins side.
will try to set this from Jenkinsfile

cdietrich:
wonder which Jenkins file is picked. the build does not pick my changes on my branch

cdietrich:
does not seem to help. next would be java 8 vs java 11.

cdietrich:
[...]

PR #512, opened 2021-02-17
Bug #508: Require tools in jenkinsfile

jonahgraham:


releng/build.Jenkinsfile
```diff
@@ -1,5 +1,9 @@
 pipeline {
   agent any
+  tools {
+    maven 'apache-maven-latest'
+    jdk 'adoptopenjdk-hotspot-jdk11-latest'
```
cdietrich, line 5: Will we move to 11 with that? Shouldn’t we use 8?

jonahgraham: Yes - I was verifying what our old machine was setup as. The PR build was explicit, but the main build was not.

PR #507, opened 2021-02-14
Fix inconsistencies in *RegistrationOptions

pisv:
Fixes #481

Signed-off-by: Vladimir Piskarev <pisv@1c.ru>

Issue #505, opened 2021-02-11
An interface-based approach to modeling data


pisv:
The LSP specification often uses multiple inheritance when describing its data structures. For example:
```typescript
export interface CallHierarchyRegistrationOptions extends
	TextDocumentRegistrationOptions, CallHierarchyOptions,
	StaticRegistrationOptions {
}
```
This is in constrast to the approach currently used in LSP4J, where `CallHierarchyRegistrationOptions` is a class, so multiple inheritance cannot be used, resulting in the following mapping:
```java
public class CallHierarchyRegistrationOptions extends AbstractTextDocumentRegistrationAndWorkDoneProgressOptions
```
Currently, `CallHierarchyRegistrationOptions` does not extend `StaticRegistrationOptions` (#481).

This can be (kind of) fixed by adding the necessary properties (`id` in this case) directly to the `CallHierarchyRegistrationOptions` class, similarly to how this is done in `ColorProviderOptions`, `FoldingRangeProviderOptions`, etc.

However, such mapping would lose an important bit of information: that `CallHierarchyRegistrationOptions` really *is a* `StaticRegistrationOptions`.

It may seem not significant for this simple example, but the underlying gap in approaches does seem quite fundamental and significant.

Leaving compatibility issues aside, it seems that an approach based on using interfaces for modeling data structures (similar to the one used by EMF) would be "a better fit" for LSP4J.

Since LSP4J is still in the 0.x version range, I thought that perhaps it would not be too late to at least start a discussion about it.

pisv:
To make it a bit more concrete, could not an Xtend declaration such as:
```java
@JsonRpcData
interface CallHierarchyRegistrationOptions extends
	TextDocumentRegistrationOptions, CallHierarchyOptions,
	StaticRegistrationOptions {
}
```
generate both a Java-interface:
```java
public interface CallHierarchyRegistrationOptions extends
	TextDocumentRegistrationOptions, CallHierarchyOptions,
	StaticRegistrationOptions {
}
```
and the corresponding implementation class:
```java
public class CallHierarchyRegistrationOptionsImpl implements
	CallHierarchyRegistrationOptions {
	// appropriate implementation: getters/setters, equals/hashCode, toString
	...
}
```
This assumes that the implementation class will be used only for instantiation. The data structure is always accessed/manipulated via the interface.

jonahgraham:
[...]

PR #515, opened 2021-02-19
Update ServerCapabilities to support work done progress

### LINKED ISSUE #480: Update ServerCapabilities to support work done progress
Some of the  Server Capabilities need to be updated to support work done progress (e.g. `ServerCapabilities.getReferencesProvider()` must return `Either<Boolean, ReferenceOptions>` instead of `Boolean`).

pisv:
Fixes #480

Signed-off-by: Vladimir Piskarev <pisv@1c.ru>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,22 @@
 
 Fixed issues: https://github.com/eclipse/lsp4j/milestone/18?closed=1
 
+Breaking API changes:
+ * Type of `ServerCapabilities.hoverProvider` changed from `Boolean` to `Either<Boolean, HoverOptions>`
+ * Type of `ServerCapabilities.definitionProvider` changed from `Boolean` to `Either<Boolean, DefinitionOptions>`
+ * Type of `ServerCapabilities.typeDefinitionProvider` changed from `Either<Boolean, StaticRegistrationOptions>` to `Either<Boolean, TypeDefinitionRegistrationOptions>`
+ * Type of `ServerCapabilities.implementationProvider` changed from `Either<Boolean, StaticRegistrationOptions>` to `Either<Boolean, ImplementationRegistrationOptions>`
+ * Type of `ServerCapabilities.referencesProvider` changed from `Boolean` to `Either<Boolean, ReferenceOptions>`
+ * Type of `ServerCapabilities.documentHighlightProvider` changed from `Boolean` to `Either<Boolean, DocumentHighlightOptions>`
+ * Type of `ServerCapabilities.documentSymbolProvider` changed from `Boolean` to `Either<Boolean, DocumentSymbolOptions>`
+ * Type of `ServerCapabilities.workspaceSymbolProvider` changed from `Boolean` to `Either<Boolean, WorkspaceSymbolOptions>`
+ * Type of `ServerCapabilities.documentFormattingProvider` changed from `Boolean` to `Either<Boolean, DocumentFormattingOptions>`
+ * Type of `ServerCapabilities.documentRangeFormattingProvider` changed from `Boolean` to `Either<Boolean, DocumentRangeFormattingOptions>`
+ * Type of `ServerCapabilities.declarationProvider` changed from `Either<Boolean, StaticRegistrationOptions>` to `Either<Boolean, DeclarationRegistrationOptions>`
+ * Type of `ServerCapabilities.callHierarchyProvider` changed from `Either<Boolean, StaticRegistrationOptions>` to `Either<Boolean, CallHierarchyRegistrationOptions>`
+ * Type of `ServerCapabilities.selectionRangeProvider` changed from `Either<Boolean, StaticRegistrationOptions>` to `Either<Boolean, SelectionRangeRegistrationOptions>`
+ * Type of `ServerCapabilities.linkedEditingRangeProvider` changed from `Either<Boolean, StaticRegistrationOptions>` to `Either<Boolean, LinkedEditingRangeRegistrationOptions>`
+
 Deprecated API changes:
  * In DAP Protocol `VariablePresentationHintKind.DATA_BREAKPOINT` has been marked @Deprecated to match its deprecation in the protocol in 1.44.
  * In DAP Protocol `SetBreakpointsArguments.lines` has been marked @Deprecated to match its deprecation in protocol many versions ago.
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -2239,6 +2239,35 @@ class CodeActionOptions extends AbstractWorkDoneProgressOptions {
 }
 
 /**
+ * Code Action registration options.
+ */
+@JsonRpcData
+class CodeActionRegistrationOptions extends AbstractTextDocumentRegistrationAndWorkDoneProgressOptions {
+	/**
+	 * CodeActionKinds that this server may return.
+	 *
+	 * The list of kinds may be generic, such as `CodeActionKind.Refactor`, or the server
+	 * may list out every specific kind they provide.
+	 */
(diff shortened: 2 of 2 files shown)


pisv:
Although I tried to be consistent in implementation, I'm not entirely satisfied with the result.

For one thing, the LSP specification itself does not seem to be always consistent when defining provider types in ServerCapabilities. Sometimes, it defines the type of a provider as

```ts
definitionProvider?: boolean | DefinitionOptions
```

and sometimes as

```ts
declarationProvider?: boolean | DeclarationOptions | DeclarationRegistrationOptions
```

Perhaps, it does not matter, since `DeclarationRegistrationOptions` extends `DeclarationOptions` and `DefinitionRegistrationOptions` extends `DefinitionOptions`, so the above definitions are probably equivalent to the following ones:

```ts
definitionProvider?: boolean | DefinitionOptions | DefinitionRegistrationOptions

declarationProvider?: boolean | DeclarationOptions
```

But I don't know whether this is really the case. Therefore, I translated it more or less literally to the following Xtend definitions:

```java
Either<Boolean, DefinitionOptions> definitionProvider

Either<Boolean, DeclarationRegistrationOptions> declarationProvider
```

and just tried to be consistent in using this approach everywhere.

Note that since `declarationProvider` was previously defined as

```java
Either<Boolean, StaticRegistrationOptions> declarationProvider
```

I just replaced `StaticRegistrationOptions` with the more strictly typed `DeclarationRegistrationOptions` rather than using a more literal (and more complex) definition:

```java
Either3<Boolean, DeclarationOptions, DeclarationRegistrationOptions> declarationProvider
```

I'm not sure whether this is entirely correct.

When working on it, I encountered an interesting case where both `ColorProviderOptions` and `FoldingRangeProviderOptions` extend `AbstractTextDocumentRegistrationAndWorkDoneProgressOptions` (i.e. are actually *RegistrationOptions) and there are no corresponding `DocumentColorOptions` and `FoldingRangeOptions` that are present in the spec (which would extend `AbstractWorkDoneProgressOptions`). That seems to be a valid alternative approach.

Also, I left `typeHierarchyProvider` as it was, since it is not specified in LSP 3.16.

Nevertheless, it is my hope that, even in its current form, it would be useful as a basis for further discussion/improvements.

jonahgraham:
IIUC this is part of what is needed to get LSP 3.16 compatibility for LSP4J 0.11.0 release. 

@spoenemann Can you review or recommend someone who can review please?

jonahgraham:
@eclipse-lsp4j-bot ok to test

KamasamaK:
- `linkedEditingRangeProvider` is still referring to `StaticRegistrationOptions` when it should be `LinkedEditingRangeRegistrationOptions`.
- `SelectionRangeRegistrationOptions` is missing its `StaticRegistrationOptions` (I missed this reviewing #507).


**Note to all implementers that the changes in `ServerCapabilities` are API breaking.**

jonahgraham:
Thanks @KamasamaK.

@pisv Please add the breaking API comments to the [changelog](https://github.com/eclipse/lsp4j/blob/master/CHANGELOG.md)

pisv:
Done. Thanks for the feedback.

KamasamaK (approved):
I agree that it's probably not necessary to have both *Options in an `Either3` since the *RegistrationOptions only add essentially optional properties. Technically, the spec has `documentSelector` required for `TextDocumentRegistrationOptions`, but it can be `null` and as long as `id` is not also defined it could just as well be the *Options. Perhaps we should add a note to `TextDocumentRegistrationOptions` that `documentSelector` is required even though it isn't `@NonNull`, but I don't think it would be serialized by default so maybe another note about how that's done.

Regarding the provider capabilities that don't use the registration options, I think it's strange but that is what the spec defines. I'm not sure why the registration options are defined, but not used. Even though the *RegistrationOptions extend the *Options in the spec, that's not true for the LSP4J classes so they wouldn't be useable in that way.

pisv:
Thank you very much for your thorough review and approval @KamasamaK

> Technically, the spec has `documentSelector` required for `TextDocumentRegistrationOptions`

IIUIC, when deserializing, this could be used to differentiate e.g. between an instance of `DeclarationRegistrationOptions` where `id` is not defined and `documentSelector` is `null`, and an instance of `DeclarationOptions`.

Then the question is whether this is necessary from the language client point of view.

I would say "no". In both cases, we would use a default document selector and a generated id. For example, see
https://github.com/microsoft/vscode-languageserver-node/blob/b90b25cc4c69824a948af71ec29e1e3c6de019d8/client/src/common/declaration.ts#L42 and https://github.com/microsoft/vscode-languageserver-node/blob/b90b25cc4c69824a948af71ec29e1e3c6de019d8/client/src/common/client.ts#L1460.

> Regarding the provider capabilities that don't use the registration options, I think it's strange but that is what the spec defines.

Perhaps we should seek a clarification from LSP guys?

> Even though the *RegistrationOptions extend the *Options in the spec, that's not true for the LSP4J classes so they wouldn't be useable in that way.

Yes, this is probably related to #505.

Thanks again.

pisv:
>> Regarding the provider capabilities...

> Perhaps we should seek a clarification...

Digging into history, in LSP 3.14 it was

```typescript
definitionProvider?: boolean;

declarationProvider?: boolean | (TextDocumentRegistrationOptions & StaticRegistrationOptions);
```

In LSP 3.15 it became

```typescript
definitionProvider?: boolean | DefinitionOptions;

declarationProvider?: boolean | DeclarationOptions | DeclarationRegistrationOptions;

export interface DefinitionOptions extends WorkDoneProgressOptions {
}

export interface DefinitionRegistrationOptions extends
	TextDocumentRegistrationOptions, DefinitionOptions {
}

export interface DeclarationOptions extends WorkDoneProgressOptions {
}

export interface DeclarationRegistrationOptions extends DeclarationOptions,
	TextDocumentRegistrationOptions, StaticRegistrationOptions  {
}
```

Clearly, this transformation was part of work done progress support. We can see a pattern here where for each server capability that had previously supported type (`TextDocumentRegistrationOptions` & `StaticRegistrationOptions`), that type was replaced with new *RegistrationOptions specific to the capability. Note also that `DeclarationRegistrationOptions` extends `StaticRegistrationOptions`, whereas `DefinitionRegistrationOptions` does not.

However, while the intent is more or less clear, its implementation is not so very clear to me. I'm not an expert in TypeScript, but since `DeclarationRegistrationOptions` extends `DeclarationOptions`, the type declaration `DeclarationOptions | DeclarationRegistrationOptions` seems redundant. And, IIUIC, since `DefinitionRegistrationOptions` extends `DefinitionOptions`, the type declaration `DefinitionOptions` implicitly also supports `DefinitionRegistrationOptions` as a valid value, although it does not seem to be the intent. Perhaps it just slipped through the cracks.

Anyway, I'm more confident in the suggested implementation now.

I'm sorry for the additional noise. A bit of paranoia because of multiple breaking changes :-)

nixel2007:
> I'm not an expert in TypeScript, but since DeclarationRegistrationOptions extends DeclarationOptions, the type declaration DeclarationOptions | DeclarationRegistrationOptions seems redundant.

I think the main idea for that declaration is something like "hey, you may pass this sniny fully functional `DeclarationRegistrationOptions` but it's okay if you'll pass only `DefinitionOptions`, lsp can handle it."

jonahgraham:
Hi folks, this issue was previously identified as being part of LSP4J 0.11.0 as it provided part of the support for completing LSP 3.16.0. If this is a non-blocking issue for 0.11.0 release, please update/remove milestone so that we can progress with the LSP4J 0.11.0 release soon. We can have a 0.12.0 release soon after the 0.11.0 release if there is value in that. In the absence of feedback I will remove the milestone in a few days so that we can release 0.11.0 as many consumers are waiting on an official LSP4J 0.11.0 release. Thank you.

pisv:
Hi Jonah, given that this PR seems to have been extensively discussed and has been approved by a committer, cannot it be just merged? It would be a pity to have the 0.11 release without this fix.

PR #517, opened 2021-03-01
Support any kind of progress in ProgressParams

### LINKED ISSUE #479: ProgressParams supports work done progress only
Currently, `ProgressParams` supports work done progress only. It does not support other kinds of progress (e.g., partial result progress).

pisv:
Fixes #479

Signed-off-by: Vladimir Piskarev <pisv@1c.ru>

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -3912,12 +3912,12 @@ class ProgressParams {
 	 */
 	@NonNull
 	@JsonAdapter(ProgressNotificationAdapter)
-	Either<? extends WorkDoneProgressNotification, ? extends JsonElement> value
+	Either<? extends WorkDoneProgressNotification, ? extends Object> value
 
 	new() {
 	}
 
-	new(@NonNull Either<String, Integer> token, @NonNull Either<? extends WorkDoneProgressNotification, ? extends JsonElement> value) {
+	new(@NonNull Either<String, Integer> token, @NonNull Either<? extends WorkDoneProgressNotification, ? extends Object> value) {
 		this.token = Preconditions.checkNotNull(token, 'token')
 		this.value = Preconditions.checkNotNull(value, 'value')
 	}
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -12,8 +12,10 @@
 package org.eclipse.lsp4j
 
 import com.google.common.annotations.Beta
+import com.google.gson.JsonElement
 import com.google.gson.annotations.JsonAdapter
 import java.util.ArrayList
+import java.util.Arrays
 import java.util.LinkedHashMap
 import java.util.List
 import java.util.Map
@@ -21,11 +23,11 @@ import org.eclipse.lsp4j.adapters.CompletionItemTextEditTypeAdapter
 import org.eclipse.lsp4j.adapters.DocumentChangeListAdapter
 import org.eclipse.lsp4j.adapters.HoverTypeAdapter
 import org.eclipse.lsp4j.adapters.InitializeParamsTypeAdapter
+import org.eclipse.lsp4j.adapters.ProgressNotificationAdapter
 import org.eclipse.lsp4j.adapters.ResourceChangeListAdapter
 import org.eclipse.lsp4j.adapters.ResourceOperationTypeAdapter
 import org.eclipse.lsp4j.adapters.SymbolInformationTypeAdapter
 import org.eclipse.lsp4j.adapters.VersionedTextDocumentIdentifierTypeAdapter
-import org.eclipse.lsp4j.adapters.WorkDoneProgressNotificationAdapter
 import org.eclipse.lsp4j.generator.JsonRpcData
 import org.eclipse.lsp4j.jsonrpc.json.adapters.JsonElementTypeAdapter
 import org.eclipse.lsp4j.jsonrpc.messages.Either
@@ -33,7 +35,6 @@ import org.eclipse.lsp4j.jsonrpc.messages.Either3
 import org.eclipse.lsp4j.jsonrpc.messages.Tuple
 import org.eclipse.lsp4j.jsonrpc.validation.NonNull
 import org.eclipse.lsp4j.util.Preconditions
-import java.util.Arrays
 
 @JsonRpcData
 class DynamicRegistrationCapabilities {
@@ -3867,13 +3868,13 @@ class ProgressParams {
 	 * The progress data.
 	 */
 	@NonNull
-	@JsonAdapter(WorkDoneProgressNotificationAdapter.Factory)
-	WorkDoneProgressNotification value
+	@JsonAdapter(ProgressNotificationAdapter)
+	Either<? extends WorkDoneProgressNotification, ? extends JsonElement> value
 
 	new() {
 	}
 
-	new(@NonNull Either<String, Integer> token, @NonNull WorkDoneProgressNotification value) {
+	new(@NonNull Either<String, Integer> token, @NonNull Either<? extends WorkDoneProgressNotification, ? extends JsonElement> value) {
 		this.token = Preconditions.checkNotNull(token, 'token')
 		this.value = Preconditions.checkNotNull(value, 'value')
 	}
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -3912,12 +3912,12 @@ class ProgressParams {
 	 */
 	@NonNull
 	@JsonAdapter(ProgressNotificationAdapter)
-	Either<? extends WorkDoneProgressNotification, ? extends Object> value
+	Either<WorkDoneProgressNotification, Object> value
 
 	new() {
 	}
 
-	new(@NonNull Either<String, Integer> token, @NonNull Either<? extends WorkDoneProgressNotification, ? extends Object> value) {
+	new(@NonNull Either<String, Integer> token, @NonNull Either<WorkDoneProgressNotification, Object> value) {
 		this.token = Preconditions.checkNotNull(token, 'token')
 		this.value = Preconditions.checkNotNull(value, 'value')
 	}
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/ProgressNotificationAdapter.java
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/ProgressNotificationAdapter.java
@@ -1,0 +1,40 @@
+/*******************************************************************************
+ * Copyright (c) 2021 1C-Soft LLC and others.
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which is available at
+ * https://www.eclipse.org/legal/epl-2.0/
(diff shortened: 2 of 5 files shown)


pisv:
I have sketched a fix for the issue. Unfortunately, I'm not an expert in Gson, so the implementation of ProgressNotificationAdapter is probably not optimal. At least, it seems to work correctly, as demonstrated by tests.

jonahgraham:
@eclipse-lsp4j-bot ok to test

I kicked off the tests with ^^^ - but I am not knowledgeable in this area to review and merge. Can a committer familiar with LSP please review.

jonahgraham:
Hi folks, this issue was previously identified as being part of LSP4J 0.11.0 as it provided part of the support for completing LSP 3.16.0. If this is a non-blocking issue for 0.11.0 release, please update/remove milestone so that we can progress with the LSP4J 0.11.0 release soon. We can have a 0.12.0 release soon after the 0.11.0 release if there is value in that. In the absence of feedback I will remove the milestone in a few days so that we can release 0.11.0 as many consumers are waiting on an official LSP4J 0.11.0 release. Thank you.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -3867,13 +3868,13 @@ class ProgressParams {
 	 * The progress data.
 	 */
 	@NonNull
-	@JsonAdapter(WorkDoneProgressNotificationAdapter.Factory)
-	WorkDoneProgressNotification value
+	@JsonAdapter(ProgressNotificationAdapter)
+	Either<? extends WorkDoneProgressNotification, ? extends JsonElement> value
```
jonahgraham, line 3872: I have a concern here that this isn't quite right as this adds Gson into the API where (AFAICT) it does not currently exist. e.g. should this be `Either<? extends WorkDoneProgressNotification, Object>`  which would more closely align it with how `Registration.registerOptions` handles this generic case using `JsonElementTypeAdapter.Factory`?

pisv: Thank you very much for review, Jonah! I fully agree with your observation and will push a fix in a minute.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -3910,13 +3911,13 @@ class ProgressParams {
 	 * The progress data.
 	 */
 	@NonNull
-	@JsonAdapter(WorkDoneProgressNotificationAdapter.Factory)
-	WorkDoneProgressNotification value
+	@JsonAdapter(ProgressNotificationAdapter)
+	Either<? extends WorkDoneProgressNotification, ? extends Object> value
```
jonahgraham, line 3915: why `? extends` - this is also inconsistent with for e.g. `SemanticTokensClientCapabilitiesRequests.range`.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/ProgressNotificationAdapter.java
```diff
@@ -0,0 +1,40 @@
...
+
+	@SuppressWarnings("unchecked")
+	@Override
+	public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
+		Gson gson2 = gson.newBuilder().registerTypeHierarchyAdapter(WorkDoneProgressNotification.class,
+				new WorkDoneProgressNotificationAdapter()).create();
+		Predicate<JsonElement> leftChecker = new PropertyChecker("kind", "begin").or(
+				new PropertyChecker("kind", "report")).or(new PropertyChecker("kind", "end"));
+		Predicate<JsonElement> rightChecker = t -> true;
+		return (TypeAdapter<T>) new EitherTypeAdapter<>(gson2, ELEMENT_TYPE, leftChecker, rightChecker);
```
jonahgraham, line 38: Does the read/write code in `JsonElementTypeAdapter` need to be used here? I don't know if it matters though as `SemanticTokensClientCapabilitiesRequests.range` does not use `JsonElementTypeAdapter`, but all cases of a field of type `Object` do, such as `InitializeParams.initializationOptions`

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
pisv: Without `? extends`, it did not compile:

```
Cannot cast from Procedures.Procedure1<JsonObject> to Procedures.Procedure1<? super Object>	JsonParseTest.java	/org.eclipse.lsp4j/src/test/xtend-gen/org/eclipse/lsp4j/test/services	line 3461	Java Problem
Cannot cast from Procedures.Procedure1<JsonObject> to Procedures.Procedure1<? super Object>	JsonSerializeTest.java	/org.eclipse.lsp4j/src/test/xtend-gen/org/eclipse/lsp4j/test/services	line 1876	Java Problem
Java problem: Cannot cast from Procedures.Procedure1<JsonObject> to Procedures.Procedure1<? super Object>	JsonParseTest.xtend	/org.eclipse.lsp4j/src/test/java/org/eclipse/lsp4j/test/services	line 1613	Xtend Problem
Java problem: Cannot cast from Procedures.Procedure1<JsonObject> to Procedures.Procedure1<? super Object>	JsonSerializeTest.xtend	/org.eclipse.lsp4j/src/test/java/org/eclipse/lsp4j/test/services	line 910	Xtend Problem
```

Probably this is an Xtend code generation issue? Not sure how to fix it.

jonahgraham: 👍 I'll have a look before merging to see what is going on here.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/ProgressNotificationAdapter.java
pisv: I'm not sure about it either. All I can tell is that when I was working on this, I was not aware about the case of `SemanticTokensClientCapabilitiesRequests.range`. Therefore, I did it similar to all cases of a field of type `Object`. Also, it seems that if it was done similar to `SemanticTokensClientCapabilitiesRequests.range`, the issue of `? extends Object` mentioned in the previous comment would probably go away.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
pisv: Thanks Jonah!

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/ProgressNotificationAdapter.java
pisv: > Also, it seems that if it was done similar to SemanticTokensClientCapabilitiesRequests.range, the issue of ? extends Object mentioned in the previous comment would probably go away.

I have tried it, and no, it would not. This is an unrelated issue, sorry.

pisv: In case I was not clear or misunderstood your comment, the current implementation of `ProgressNotificationAdapter` will parse a value such as `"value": { "foo": "bar" }` into a `JsonObject` rather than a `LinkedHashMap`. I thought it was correct because I had not known about the case of `SemanticTokensClientCapabilitiesRequests.range`. If it would be better to parse it in a `LinkedHashMap` (e.g., for consistency with `SemanticTokensClientCapabilitiesRequests.range`), please let me know and I'll fix it.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
jonahgraham: The solution to removing the `? extends` is like this in the .xtend file:

```java
			params = new ProgressParams => [
				token = Either.forRight(1234)
				value = Either.forRight((new JsonObject => [
					addProperty("foo", "bar")
				]) as Object)
			]
```

i.e. upcast the `JsonObject` as `Object`. 

or specify upfront which types in the Either generics we want:

```java
			params = new ProgressParams => [
				token = Either.forRight(1234)
				value = Either.<WorkDoneProgressNotification, Object>forRight(new JsonObject => [
					addProperty("foo", "bar")
				])
			]
```

I do think there is an argument to be made that all the non-final types should be `? extends` in Protocol, but I think for now consistency is better and if you think it should be `? extends` then we should handle that in a new issue.

WDYT? If you like one of the above solutions, can you update the PR.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/ProgressNotificationAdapter.java
jonahgraham: I vote to leave it as you have written it now - it is consistent with other uses. If it is wrong, we should fix all the cases in a new issue.

jonahgraham: > In case I was not clear or misunderstood your comment, the current implementation of `ProgressNotificationAdapter` will parse a value such as `"value": { "foo": "bar" }` into a `JsonObject` rather than a `LinkedHashMap`. I thought it was correct because I had not known about the case of `SemanticTokensClientCapabilitiesRequests.range`. If it would be better to parse it in a `LinkedHashMap` (e.g., for consistency with `SemanticTokensClientCapabilitiesRequests.range`), please let me know and I'll fix it.

I didn't see this comment before I just marked as resolved. Thanks for looking into this more are identifying this difference. I think it should be made consistent and where possible not have `JsonObject`s being returned from the LSP4J API. I don't know if that is a rule or design intention in the past though.

pisv: > I think it should be made consistent and where possible not have JsonObjects being returned from the LSP4J API. I don't know if that is a rule or design intention in the past though.

I must say I'm a bit puzzled now. I have found issue #126, where `JsonElementTypeAdapter` was introduced specifically for 'parsing Object into a JsonElement'. It looks like the case of `SemanticTokensClientCapabilitiesRequests.range` is an exception rather than a rule. Therefore, I would rather concur with your previous suggestion:

> I vote to leave it as you have written it now - it is consistent with other uses.

WDYT?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
pisv: Done!

> I do think there is an argument to be made that all the non-final types should be ? extends in Protocol, but I think for now consistency is better and if you think it should be ? extends then we should handle that in a new issue.

I fully agree.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/ProgressNotificationAdapter.java
jonahgraham: > I think it should be made consistent and where possible not have JsonObjects being returned from the LSP4J API. I don't know if that is a rule or design intention in the past though.

Yeah - so that wasn't true. Thank you for looking a little deeper and pointing #126.

jonahgraham: >> I vote to leave it as you have written it now - it is consistent with other uses.
> WDYT?

Yes.

PR #520, opened 2021-03-13
Remove unused Java file

KamasamaK:
I believe this was used in an earlier iteration of the Call Hierarchy API, but it has no use anymore.


jonahgraham:
This doesn't seem to be needed for 0.11.0 so deferring. It can go in right after I release later today, but needs a changelog entry too.

KamasamaK:
Yes, this is not necessary for 0.11.0. If it's actually unused as I suspect, I don't think it warrants a change log entry. I think it was just overlooked in #387.

PR #523, opened 2021-03-22
Fix #522: Change the field type 'matches' of 'FileOperationPattern' to be String

### LINKED ISSUE #522: No way to call FileOperationPattern.setMatches() to only match files or folders
When i want to register `FileOperationsServerCapabilities` to enable willRenameFiles support, i found it's not possible to create a FileOperationPattern to only match files or folders.

The setter of the field `FileOperationPatternKind matches` asks for an instance of FileOperationPatternKind. See https://github.com/eclipse/lsp4j/blob/4901031aa9fb277711b68a2a8dc0a234fab41b5f/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend#L7036-L7058

But the class `FileOperationPatternKind` hides its constructor and no way to create an instance for it. See 
https://github.com/eclipse/lsp4j/blob/4901031aa9fb277711b68a2a8dc0a234fab41b5f/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/FileOperationPatternKind.java#L20-L33

We should make the matches field to be String type, and accept FileOperationPatternKind.File and FileOperationPatternKind.Folder as value.

A fix could be:
```
class FileOperationPattern {

    	/**
	 * Whether to match files or folders with this pattern.
	 *
	 * The match kind could be <code>FileOperationPatternKind.File</code>, <code>FileOperationPatternKind.Folder</code>.
	 *
	 * Matches both if undefined.
	 */
	String matches

}
```

testforstephen:
Signed-off-by: Jinbo Wang <jinbwan@microsoft.com>

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -7053,10 +7053,9 @@ class FileOperationPattern {
 	/**
 	 * Whether to match files or folders with this pattern.
 	 *
-	 * The match kind could be <code>FileOperationPatternKind.File</code>,
-	 * <code>FileOperationPatternKind.Folder</code> and undefined.
-	 *
 	 * Matches both if undefined.
+	 *
+	 * See {@link FileOperationPatternKind} for allowed values.
 	 */
 	String matches
 
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -7053,9 +7053,12 @@ class FileOperationPattern {
 	/**
 	 * Whether to match files or folders with this pattern.
 	 *
+	 * The match kind could be <code>FileOperationPatternKind.File</code>,
+	 * <code>FileOperationPatternKind.Folder</code> and undefined.
+	 *
 	 * Matches both if undefined.
 	 */
-	FileOperationPatternKind matches
+	String matches
 
 	/**
 	 * Additional options used during matching.


KamasamaK:
@eclipse-lsp4j-bot ok to test

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -7053,9 +7053,12 @@ class FileOperationPattern {
 	/**
 	 * Whether to match files or folders with this pattern.
 	 *
+	 * The match kind could be <code>FileOperationPatternKind.File</code>,
+	 * <code>FileOperationPatternKind.Folder</code> and undefined.
+	 *
 	 * Matches both if undefined.
 	 */
-	FileOperationPatternKind matches
```
cdietrich, line 7058: is this a general new pattern how we deal with enums?

cdietrich: should we also remove the enum then?

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -7053,9 +7053,12 @@ class FileOperationPattern {
 	/**
 	 * Whether to match files or folders with this pattern.
 	 *
+	 * The match kind could be <code>FileOperationPatternKind.File</code>,
```
KamasamaK, line 7056: Use `{@link FileOperationPatternKind}` to reference the source of values.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
KamasamaK: Declaring it as a `String` is the standard pattern used in this project when the enum is a string. The *Kind class is just used as a helper in this case. It was simply a mistake to not make it a `String` initially.

KamasamaK: See #283

testforstephen: I just quickly search the classes with `*Kind`, looks like we use enum for number type (such as `CompletionItemKind`), and string literal for string type (such as `FoldingRangeKind`).

`FileOperationPatternKind` accepts `file` and `folder` as value, so using string literals look ok.

testforstephen: thank @KamasamaK for clarification.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
testforstephen: Nice suggestion!

KamasamaK (approved):

PR #530, opened 2021-03-26
fixed some warnings

cdietrich:
fixed some warnings

### DIFF
--- a/org.eclipse.lsp4j.jsonrpc.debug/src/main/java/org/eclipse/lsp4j/jsonrpc/debug/adapters/DebugMessageTypeAdapter.java
+++ b/org.eclipse.lsp4j.jsonrpc.debug/src/main/java/org/eclipse/lsp4j/jsonrpc/debug/adapters/DebugMessageTypeAdapter.java
@@ -295,7 +295,7 @@ public class DebugMessageTypeAdapter extends MessageTypeAdapter {
 		} else if ("response".equals(messageType) && success != null && success) {
 			return super.parseResult(in, Integer.toString(request_seq));
 		} else {
-			return new JsonParser().parse(in);
+			return JsonParser.parseReader(in);
 		}
 	}
 
--- a/releng/lsp4j-feature/feature.xml
+++ b/releng/lsp4j-feature/feature.xml
@@ -26,7 +26,7 @@ SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
    </license>
 
    <requires>
-      <import plugin="com.google.gson" version="2.8.0" match="equivalent"/>
+      <import plugin="com.google.gson" version="2.8.6" match="equivalent"/>
       <import plugin="org.eclipse.xtend.lib" version="2.10.0" match="compatible"/>
    </requires>
 
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -16,7 +16,7 @@ ext.versions = [
 	'xtend_lib': '2.24.0',
 	'guava': '[14.0,31)',
 	'guava_orbit': '30.1.0.v20210127-2300',
-	'gson': '2.8.6',
+	'gson': '[2.8.6,2.9)',
 	'gson_orbit': '2.8.6.v20201231-1626',
 	'websocket_jakarta': '2.0.0',
 	'junit': '4.12'
--- a/org.eclipse.lsp4j.debug/build.gradle
+++ b/org.eclipse.lsp4j.debug/build.gradle
@@ -25,5 +25,6 @@ jar {
 
 jar.manifest {
 	instruction 'Import-Package', "com.google.common.*;version=\"$versions.guava\""
+	instruction 'Import-Package', "com.google.gson.*;version=\"$versions.gson\""
 	instruction 'Import-Package', '*'
 }
--- a/org.eclipse.lsp4j.generator/build.gradle
+++ b/org.eclipse.lsp4j.generator/build.gradle
@@ -20,5 +20,6 @@ dependencies {
 
 jar.manifest {
 	instruction 'Import-Package', "com.google.common.*;version=\"$versions.guava\""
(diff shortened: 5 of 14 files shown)


pisv:
Thanks for the clean-up, Christian!

org.eclipse.lsp4j.jsonrpc.debug/src/main/java/org/eclipse/lsp4j/jsonrpc/debug/adapters/DebugMessageTypeAdapter.java
```diff
@@ -295,7 +295,7 @@ protected Object parseBody(JsonReader in, String messageType, int request_seq, S
 		} else if ("response".equals(messageType) && success != null && success) {
 			return super.parseResult(in, Integer.toString(request_seq));
 		} else {
-			return new JsonParser().parse(in);
+			return JsonParser.parseReader(in);
```
jonahgraham, line 298: This change needs a corresponding change in the manifest ranges for Gson. Gson does not (strictly) follow Semantic versioning and this new method (and the deprecation of what was removed) was introduced in 2.8.6. Our current code reads:

```
Import-Package: com.google.gson;version="[2.8,3)"
```

However, I don't know what magic generates that range.

Note that the Eclipse platform had this issue and it was fixed just before 2021-03 release https://bugs.eclipse.org/bugs/show_bug.cgi?id=571550

This applies to many of the changes below, I am not marking each.

jonahgraham (changes requested):

org.eclipse.lsp4j.jsonrpc.debug/src/main/java/org/eclipse/lsp4j/jsonrpc/debug/adapters/DebugMessageTypeAdapter.java
cdietrich: good point. the only reference i found was the 2.8.6

cdietrich: wonder if this comes from the old gradle osgi plugin

jonahgraham: Adding this above existing instuction is close - but not quite right in https://github.com/eclipse/lsp4j/blob/ad99a3d77139351dbc00c2dacd388fe749652309/org.eclipse.lsp4j.jsonrpc.debug/build.gradle#L22-L24
```
	instruction 'Import-Package', "com.google.gson.*;version=\"$versions.gson\""
```

For some reason the above causes downstream bundles in gradle to fail to build for me, and the above is a lower bound, but there should be an upper bound. The above generates:

```
Import-Package: com.google.gson;version="2.8.6",
```

cdietrich: something like for guava? do we assume nothing breaks with 2.9 etc?

jonahgraham: Yes I think you are right, this needs updating too:

https://github.com/eclipse/lsp4j/blob/da194124f75c8b18be15a3c9cab03f6b62fdb92a/gradle/versions.gradle#L19

like 

https://github.com/eclipse/lsp4j/blob/da194124f75c8b18be15a3c9cab03f6b62fdb92a/gradle/versions.gradle#L17

cdietrich: the question is: should the upper bound be 2.9 or 3?

releng/lsp4j-feature/feature.xml
```diff
@@ -26,7 +26,7 @@ SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
    </license>
 
    <requires>
-      <import plugin="com.google.gson" version="2.8.0" match="equivalent"/>
+      <import plugin="com.google.gson" version="2.8.6" match="equivalent"/>
```
jonahgraham, line 29: thanks for catching this part

org.eclipse.lsp4j.jsonrpc.debug/src/main/java/org/eclipse/lsp4j/jsonrpc/debug/adapters/DebugMessageTypeAdapter.java
jonahgraham: We are saying the same thing at the same time :-) I think this is resolved now with your version fixes.

jonahgraham (approved):

cdietrich:
is org.eclipse.lsp4j.jsonrpc.test.IntegrationTest.testMalformedJson2 flaky?

cdietrich:
@eclipse-lsp4j-bot run tests

jonahgraham:
> is org.eclipse.lsp4j.jsonrpc.test.IntegrationTest.testMalformedJson2 flaky?

I don't think so - passing consistently on master https://ci.eclipse.org/lsp4j/job/lsp4j-multi-build/job/master/

jonahgraham:
> > is org.eclipse.lsp4j.jsonrpc.test.IntegrationTest.testMalformedJson2 flaky?
> 
> I don't think so - passing consistently on master https://ci.eclipse.org/lsp4j/job/lsp4j-multi-build/job/master/

Well - it may pass a lot, but the test does have a race condition. See #531

PR #532, opened 2021-03-26
Fixes #531: Remove race condition from integration tests

### LINKED ISSUE #531: org.eclipse.lsp4j.jsonrpc.test.IntegrationTest.testMalformedJson2 is flaky
The test has a race condition.

jonahgraham:
Some of the tests were checking for output which is produced on another
thread but with no synchronization to make sure the output has
finished being written by that thread.

By adding the "get" to startListening the test does not process
the output (or log message) until the other end has shutdown.

The reason this worked of the time is that there was partial synchronization
by having the logMessages.await - it caused the main thread to wait
long enough most of the time. However as the log messages are generated
before the tested output, there was still a small window for failure.

This was tested by adding a Thread.sleep(100) to StreamMessageConsumer.consume
and StreamMessageProducer.listen to check if the tests still passed.

A sleep longer than a few hundred milliseconds causes the tests to simply
timeout.


pisv:
Thanks for the fix Jonah! This looks fine to me. I just noticed a couple of typos in the thoroughly written commit message. I've amended them in the first comment on the PR, hoping you would not mind. The commit message itself can be trivially fixed at merge time, of course.

jonahgraham:
Thanks @pisv for the review and typo fixes!

PR #535, opened 2021-03-29
Improve Javadoc links

KamasamaK:
Replace Markdown code/links with Javadoc links.


jonahgraham:
LGTM

pisv:
Thanks @KamasamaK for this great improvement. LGTM too.

Although not exactly related to the aim of this PR, it seems that the Javadoc for `FileOperationPattern.glob` could still be improved. Currently, it is rendered as a single line in Javadoc viewer. Of course, it can also be addressed in a separate PR. (In which case I can try to handle it, if you don't mind.)

pisv:
@KamasamaK Yes, it would be better that way.

PR #543, opened 2021-04-06
Missing fields

KamasamaK:
Adds some error code changes from 3.16 and earlier. Also adds `allCommitCharacters` from 3.2.

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -6642,6 +6642,19 @@ class CompletionRegistrationOptions extends AbstractTextDocumentRegistrationAndW
 	 */
 	Boolean resolveProvider
 
+	/**
+	 * The list of all possible characters that commit a completion. This field
+	 * can be used if clients don't support individual commit characters per
+	 * completion item. See client capability
+	 * {@link CompletionItemCapabilities#commitCharactersSupport}.
+	 *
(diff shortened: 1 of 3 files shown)


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -2534,6 +2534,19 @@ class CompletionOptions extends AbstractWorkDoneProgressOptions {
...
+	 * can be used if clients don't support individual commit characters per
+	 * completion item. See client capability
+	 * {@link CompletionItemCapabilities#commitCharactersSupport}.
+	 *
+	 * If a server provides both {@code allCommitCharacters} and commit characters on
+	 * an individual completion item the ones on the completion item win.
+	 *
+	 * Since 3.2.0
+	 */
+	List<String> allCommitCharacters
```
pisv, line 2548: Also needs to be added to `CompletionRegistrationOptions`. Another instance of the problem discussed in #505.

KamasamaK: This also involves what was discussed in #515 where it was pointed that that some *RegistrationOptions like `CompletionRegistrationOptions` are unused.

pisv: Just a note that, although `CompletionRegistrationOptions` is not used as part of the return type for `completionProvider`, it is specified to be used for registration of the corresponding dynamic capability.

pisv (approved):

PR #546, opened 2021-04-13
Add Either.map(...)

mickaelistria:
Many case of usage of Either in LSP4E involve turning Either into a
value. That requires some boilerplate: declare variable, if/else.

This constructs can allow for instance replacing

	Either<TextEdit, InsertReplaceEdit> textEdit = item.getTextEdit();
	if (textEdit.isLeft())
		return offset ==
LSPEclipseUtils.toOffset(textEdit.getLeft().getRange().getStart(),
document);
	else {
		Position replace = textEdit.getRight().getReplace().getStart();
		Position insert = textEdit.getRight().getInsert().getStart();
		Position start = replace;
		if (replace.getLine() > insert.getLine() || replace.getCharacter() >
insert.getCharacter()) start = insert;
			return offset == LSPEclipseUtils.toOffset(start, document);
		}

by

	return offset == LSPEclipseUtils.toOffset(
		textEdit.map(
			edit -> edit.getRange().getStart(),
			insertReplaceEdit -> {
				Position replace = insertReplaceEdit.getReplace().getStart();
				Position insert = insertReplaceEdit.getInsert().getStart();
				Position start = replace;
				if (replace.getLine() > insert.getLine() || replace.getCharacter() >
insert.getCharacter()) start = insert;
			}), document);

which better captures the logic.

### DIFF
--- a/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/messages/Either.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/messages/Either.java
@@ -15,6 +15,7 @@ import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.function.Function;
 
 import org.eclipse.lsp4j.jsonrpc.validation.NonNull;
 
@@ -24,11 +25,11 @@ import org.eclipse.lsp4j.jsonrpc.validation.NonNull;
 public class Either<L, R> {
 
 	public static <L, R> Either<L, R> forLeft(@NonNull L left) {
-		return new Either<L, R>(left, null);
+		return new Either<>(left, null);
 	}
 
 	public static <L, R> Either<L, R> forRight(@NonNull R right) {
-		return new Either<L, R>(null, right);
+		return new Either<>(null, right);
 	}
 
 	private final L left;
@@ -63,7 +64,17 @@ public class Either<L, R> {
 	public boolean isRight() {
 		return right != null;
 	}
-	
+
+	public <T> T map(@NonNull Function<? super L, ? extends T> mapLeft, @NonNull Function <? super R, ? extends T> mapRight) {
+		if (isLeft()) {
+			return mapLeft.apply(getLeft());
+		}
+		if (isRight()) {
+			return mapRight.apply(getRight());
+		}
+		return null;
+	}
+
 	@Override
 	public boolean equals(Object obj) {
 		if (obj instanceof Either<?, ?>) {
--- a/org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/json/EitherTest.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/json/EitherTest.java
@@ -11,6 +11,10 @@
  ******************************************************************************/
 package org.eclipse.lsp4j.jsonrpc.test.json;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
 import java.util.Objects;
 
 import org.eclipse.lsp4j.jsonrpc.json.adapters.EitherTypeAdapter;
@@ -22,9 +26,6 @@ import org.junit.Test;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonObject;
-
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
 
 
 public class EitherTest {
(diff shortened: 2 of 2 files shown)


mickaelistria:
cc @jcompagner

jonahgraham:
Lgtm - could we get a test for the new code? There are some other tests for either here:

https://github.com/eclipse/lsp4j/blob/1470d748e556476369dfaf6a77cb32efddb764be/org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/json/EitherTest.java#L79-L90

mickaelistria:
> could we get a test for the new code? 

Done.

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/messages/Either.java
```diff
@@ -63,7 +64,17 @@ public boolean isLeft() {
 	public boolean isRight() {
 		return right != null;
 	}
-	
+
+	public <T> T map(Function<L, T> mapLeft, Function <R, T> mapRight) {
```
szarnekow, line 68: The signature should be `map(Function<? super L, ? extends T> mapLeft, Function<? super R, ? extends T> mapRight)` I suppose. Also the functions should be mandatory and null mappers should be prohibited to avoid surprises.

pisv:
Just a note that there is also `Either3`, where a three-argument overload for `map` would probably make sense.

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/messages/Either.java
mickaelistria: Thanks for the suggestion, I applied it.

mickaelistria:
Anything else I should do to complete this PR?

szarnekow:
> Anything else I should do to complete this PR?

What do you think about @pisv remark?

> Just a note that there is also Either3, where a three-argument overload for map would probably make sense.

mickaelistria:
> What do you think about @pisv remark?

I think it's right, but I don't plan to work on this case right now as I don't see usage of Either3 in LSP4E for the moment. So that would be another patch later maybe.

pisv:
> > What do you think about @pisv remark?
> 
> I think it's right, but I don't plan to work on this case right now as I don't see usage of Either3 in LSP4E for the moment. So that would be another patch later maybe.

No worries. For the sake of API consistency, I'll take care of it after this PR is merged. Thank you for the contribution. (+1 from my side.)

pisv:
@eclipse-lsp4j-bot ok to test

pisv:
@mickaelistria I've also tried to run the build for this PR locally (using `./gradlew build`) and it fails with `EitherTest.java:16: error: package org.junit.jupiter.api does not exist`. Can you please take a look at it?

jcompagner:
is that not a matter of changing this import
import static org.junit.jupiter.api.Assertions.assertEquals;

to just 

import static org.junit.Assert..assertEquals;

pisv:
@eclipse-lsp4j-bot run tests

pisv:
@jonahgraham The local build completes successfully now. Not sure what happens with our Jenkins PR Build. Strangely, the test result does not even include tests in the `org.eclipse.lsp4j.jsonrpc.test.json` package. Can you please help?

org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/json/EitherTest.java
```diff
@@ -107,6 +110,19 @@ public void testEqualsFalseWithNonNull() {
 		assertFalse(either1.equals(either2));
 	}
 
+	@Test
+	public void testMap() {
+		Collection<Either<Collection<?>, String>> toTest = List.of(
```
jonahgraham, line 115: LSP4J is Java 8 - No List.of available:

```
07:31:37 > Task :org.eclipse.lsp4j.jsonrpc:compileTestJava FAILED
07:31:37 /home/jenkins/agent/workspace/lsp4j-github-pullrequests/org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/json/EitherTest.java:116: error: cannot find symbol
07:31:37 			Either.forLeft(List.of("f", "o", "o")),
07:31:37 			                   ^
07:31:37   symbol:   method of(String,String,String)
07:31:37   location: interface List
07:31:37 1 error
```

jonahgraham:
> @jonahgraham The local build completes successfully now. Not sure what happens with our Jenkins PR Build. Strangely, the test result does not even include tests in the `org.eclipse.lsp4j.jsonrpc.test.json` package. Can you please help?

I suspect you are using Java 11 locally? The build machine uses Java 8 as there is a Java 8 requirement. I don't know whether there is an actual requirement to stay on Java 8 or not.

pisv:
> I suspect you are using Java 11 locally? The build machine uses Java 8 as there is a Java 8 requirement. I don't know whether there is an actual requirement to stay on Java 8 or not.

Yes. Got it. Thanks for the clarification and your help!

mickaelistria:
I tried to Java8-ify this test, but I would really appreciate if LSP4J can move to Java 11 in a near future. Java 8 is really a burden that doesn't make sense to keep bothering with (unless LSP4J has some major contributor really needing this and contributing enough to balance the extra effort).

jonahgraham:
> Java 8 is really a burden that doesn't make sense to keep bothering with (unless LSP4J has some major contributor really needing this and contributing enough to balance the extra effort).

I just created #547 - we should at least know why we are still on Java 8.

PR #550, opened 2021-04-21
Add three-argument overload for map to Either3

pisv:
This is a follow-up to #546

### DIFF
--- a/org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/json/EitherTest.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/json/EitherTest.java
@@ -16,6 +16,7 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
 import java.util.Objects;
+import java.util.function.Function;
 
 import org.eclipse.lsp4j.jsonrpc.json.adapters.EitherTypeAdapter;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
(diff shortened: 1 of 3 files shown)


pisv:
I'm going to merge it later today if there are no comments.

org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/json/EitherTest.java
```diff
@@ -110,16 +111,33 @@ public void testEqualsFalseWithNonNull() {
...
+		either = Either.forRight("bar");
+		assertEquals("BAR", either.map(
 				String::new,
 				String::toUpperCase));
 	}
 
+	@Test
+	public void testMapEither3() {
+		Either3<String, Integer, Boolean> either3;
+		Function<String, String> mapFirst = s -> s.toUpperCase();
```
jonahgraham, line 127: Only the most minor of style comments - the `testMap` above uses `String::toUppercase` and this code uses `s -> s.toUpperCase()`

jonahgraham (approved):
lgtm

org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/json/EitherTest.java
pisv: Thank you for the review, Jonah! Yes, not using a method reference here is (kind of) intentional, to better match the following definitions of `mapSecond` and `mapThird`, where method references unfortunately cannot be used because of the ambiguity reported by the compiler. So, I would suggest to either leave this code as it is or change the code in `testMap` to also not use method references for more consistency (I slightly prefer the first alternative). WDYT?

jonahgraham: Leave the code as is.

pisv: OK. Thanks!

PR #552, opened 2021-05-06
Improve Javadoc formatting in Protocol.xtend

pisv:
Add missing p tags, replace Markdown lists with HTML lists, etc.


pisv:
If this change is welcome, I'd like to get it in sooner rather than later to avoid potential merge conflicts down the road. Please let me know if there are any objections or comments.

jonahgraham:
I think this is a fine improvement - @pisv feel free to merge this, you have given plenty of time for anyone who thinks it is a bad idea to say so.

PR #529, opened 2021-03-26
[#528] remove runtime dependencies to xbase.lib

cdietrich:
[#528] remove some runtime dependencies to xbase.lib
Signed-off-by: Christian Dietrich <christian.dietrich@itemis.de>

Issue #528, opened 2021-03-26
Deal with Xtext leaving the simrel


cdietrich:
Xtext plans to leave eclipse simrel
https://github.com/eclipse/xtext/issues/1961

we currently consume xbase.lib and xtend.lib at runtime from them.

cdietrich:
Besides Pure and ToStringBuilder https://github.com/eclipse/lsp4j/issues/494
there are calls to collectionliterals

PR #557, opened 2021-06-16
repaired xtend build. workaround for eclipse/xtext#1976

cdietrich:
repaired xtend build. workaround for eclipse/xtext#1976

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,6 +39,15 @@ subprojects {
 	apply from: "${rootDir}/gradle/xtend-compiler-settings.gradle"
 	apply from: "${rootDir}/gradle/maven-deployment.gradle"
 	apply from: "${rootDir}/gradle/manifest-gen.gradle"
+
+	configurations.all {
+		if (name.contains("Tooling")) {
+			dependencies {
+				add(name, "org.eclipse.xtend:org.eclipse.xtend.core:${versions.xtend_lib}")
+				add(name, platform("org.eclipse.xtext:xtext-dev-bom:${versions.xtend_lib}"))
+			}
+		}
+	}
 }
 
 task clean(type: Delete) {


build.gradle
```diff
@@ -39,8 +39,24 @@ subprojects {
 	apply from: "${rootDir}/gradle/xtend-compiler-settings.gradle"
 	apply from: "${rootDir}/gradle/maven-deployment.gradle"
 	apply from: "${rootDir}/gradle/manifest-gen.gradle"
+
+	configurations.all {
+		if (name.endsWith("ToolingMain") || name.endsWith("ToolingTest")) {
```
oehme, line 44: Why not `contains("Tooling")`?

cdietrich: should also work.
unfortunately this workaround that works local, does not work on server

cdietrich: now it seems to work fine

build.gradle
```diff
@@ -39,6 +39,21 @@ subprojects {
 	apply from: "${rootDir}/gradle/xtend-compiler-settings.gradle"
 	apply from: "${rootDir}/gradle/maven-deployment.gradle"
 	apply from: "${rootDir}/gradle/manifest-gen.gradle"
+
+	configurations.all {
+		if (name.contains("Tooling")) {
+			resolutionStrategy { 
+				eachDependency { DependencyResolveDetails details ->
+					if (details.requested.group == 'org.eclipse.platform' && details.requested.name == 'org.eclipse.core.runtime') {
+						details.useVersion "3.19.0"
```
jonahgraham, line 48: Is the syntax difference here to line 51 `(` and `)` important.

jonahgraham (changes requested):
@cdietrich I am pleased that this makes the build work - but from the commit comment I don't understand why it was broken, nor why this can't be solved by adding versions to https://github.com/eclipse/lsp4j/blob/85273590aea17a136e37a633a777682c3f36c3e4/gradle/versions.gradle#L15-L18 ? I have read through the extremely long discussions you have had to try to figure this out in https://github.com/eclipse/xtext/issues/1976

Can you please add a summary/comment somewhere so that this can be maintained going forward.

IIUC something like:

> We need to keep some bundles on the last version that works with Java 8 so we can continue to build and target Java 8 with LSP4J.

Additionally, are these dependencies build time only, or under what circumstance does it become a runtime dep too? Does this relate to the removal of runtime deps on xbase in #528 and #529?

Thanks.

cdietrich:
summary. platform and jdt have version ranges. these version ranges are considered by gradle to bee good and that its a good idea to pull the latest greatest core.runtime and equinix.common versions. these run with java 11 only but the build here runs with java 8, so this will explode. 

so here i fix it by telling gradle to pull the versions are are intended to be used by gradle.

an alternative solution would be to run the build with java 11 against java 8 target

build.gradle
cdietrich: nope they are not, will harmonize

jonahgraham (approved):
Thanks @cdietrich for the summary.

build.gradle
```diff
@@ -39,6 +39,21 @@ subprojects {
 	apply from: "${rootDir}/gradle/xtend-compiler-settings.gradle"
 	apply from: "${rootDir}/gradle/maven-deployment.gradle"
 	apply from: "${rootDir}/gradle/manifest-gen.gradle"
+
+	configurations.all {
+		if (name.contains("Tooling")) {
+			resolutionStrategy { 
+				eachDependency { DependencyResolveDetails details ->
+					if (details.requested.group == 'org.eclipse.platform' && details.requested.name == 'org.eclipse.core.runtime') {
```
oehme, line 47: Can you add the xtext-dev-bom here instead of a resolution strategy?

cdietrich: Will give it a try tomorrow

cdietrich: something like https://github.com/eclipse/lsp4j/compare/cd_repairBuild_variantB @oehme ?

oehme: Yep :+1:

cdietrich: updated the PR

jonahgraham (approved):
I simply don't know gradle well enough - but this looks nice and clean and I like how in this revision the hard coded version numbers are gone.

cdietrich:
@jonahgraham  you can find them here https://github.com/eclipse/xtext-lib/blob/d8f2aba7be9b515cf53e31d820d2f8bcd00de97d/org.eclipse.xtext.dev-bom/build.gradle#L64

jonahgraham:
> @jonahgraham you can find them here https://github.com/eclipse/xtext-lib/blob/d8f2aba7be9b515cf53e31d820d2f8bcd00de97d/org.eclipse.xtext.dev-bom/build.gradle#L64

That is a good place for them :-) thanks for the link

PR #559, opened 2021-07-03
typo in documentation

fredg1:
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#initialize

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/LanguageServer.java
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/LanguageServer.java
@@ -32,7 +32,7 @@ public interface LanguageServer {
 	 * the server.
 	 * 
 	 * If the server receives request or notification before the initialize request it should act as follows:
-	 * 	- for a request the respond should be errored with code: -32001. The message can be picked by the server.
+	 * 	- for a request the respond should be errored with code: -32002. The message can be picked by the server.
 	 *  - notifications should be dropped, except for the exit notification. This will allow the exit a server without an initialize request.
 	 *  
 	 * Until the server has responded to the initialize request with an InitializeResult 
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/LanguageServer.java
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/LanguageServer.java
@@ -39,7 +39,7 @@ public interface LanguageServer {
 	 *    This will allow the client to exit a server without an initialize request.
 	 *  
 	 * Until the server has responded to the initialize request with an InitializeResult,
-	 * the client must not sent any additional requests or notifications to the server.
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/LanguageServer.java
```diff
@@ -32,7 +32,7 @@
 	 * the server.
 	 * 
 	 * If the server receives request or notification before the initialize request it should act as follows:
-	 * 	- for a request the respond should be errored with code: -32001. The message can be picked by the server.
+	 * 	- for a request the respond should be errored with code: -32002. The message can be picked by the server.
```
pisv, line 35: Thanks for the catch! Just a couple of suggestions for potential enhancements:
- There is the corresponding error code constant, `ResponseErrorCode.serverNotInitialized`, which could be linked to in the javadoc instead of embedding the value.
- There is another typo in this sentence: `respond` -> `response`.

KamasamaK: I agree. It should be `{@link org.eclipse.lsp4j.jsonrpc.messages.ResponseErrorCode#serverNotInitialized}` instead. Also, it should be "If the server receives **_a_** request or notification...". There are going to be many typos like this because this was copied from an old version of [the spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/), which often needs to be corrected because the primary maintainer does not have strong English skills.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/LanguageServer.java
```diff
@@ -31,24 +31,27 @@
...
+	 * If the server receives requests or notifications before the initialize request, it should act as follows:
+	 *  - for a request, the response should be errored with:
+	 *    {@link org.eclipse.lsp4j.jsonrpc.messages.ResponseErrorCode#serverNotInitialized}.
+	 *    The message can be picked by the server.
+	 *  - notifications should be dropped, except for the exit notification.
+	 *    This will allow the client to exit a server without an initialize request.
 	 *  
-	 * Until the server has responded to the initialize request with an InitializeResult 
+	 * Until the server has responded to the initialize request with an InitializeResult,
 	 * the client must not sent any additional requests or notifications to the server.
```
pisv, line 42:
```suggestion
	 * the client must not send any additional requests or notifications to the server.
```

pisv (commented):
Thanks for going the extra mile in improving the documentation! LGTM.

Perhaps there is still some room for enhancement, like adding some `{@link}` or `{@code}` tags as well as translating markdown lists/line breaks to HTML, but personally I'm fine with accepting the patch as is, because these markup-related notes are unrelated to fixing the typos.

pisv:
@eclipse-lsp4j-bot ok to test

KamasamaK (approved):
LGTM

@pisv Feel free to merge if it looks good to you.

PR #561, opened 2021-07-04
proposed documentation of error generation

fredg1:


documentation/jsonrpc.md
```diff
@@ -29,11 +29,28 @@ You always work with two `Endpoints`. Usually one of the endpoints, a `RemoteEnd
...
 
-For requests the story is slightly more complicated. When a request message comes in, the `RemoteEndpoint` tracks the request `id` and invokes `request` on the local endpoint. In addition it adds completion stage to the returned `CompletableFuture`, that translates the result into a JSON RPC response message.
+For requests, the story is slightly more complicated. When a request message comes in, the `RemoteEndpoint` tracks the request `id` and invokes `request` on the local endpoint. In addition it adds completion stage to the returned `CompletableFuture`, that translates the result into a JSON RPC response message.
 
 For the other direction, if the implementation calls request on the RemoteEndpoint, the message is send and tracked locally. 
 The returned `CompletableFuture` will complete once a corresponsing result message is received.
 
+## Errors
+
+If requests are malformed, or for any other reason laid out by the LSP, a `ResponseError` must be sent instead.
```
pisv, line 39:
```suggestion
The receiver of a request always needs to return a response message to conform to the JSON RPC specification. In case the result value cannot be provided in a response because of an error, the `error` property of the `ResponseMessage` must be set to a `ResponseError` describing the failure.
```

documentation/jsonrpc.md
```diff
@@ -29,11 +29,28 @@ You always work with two `Endpoints`. Usually one of the endpoints, a `RemoteEnd
...
+
+This is done by throwing a `ResponseErrorException`, which carries the `ResponseError` to attach to the response. Simple example:
```
pisv, line 41:
```suggestion
This can be done by throwing a `ResponseErrorException` from the request message handler in a local endpoint. The exception carries a `ResponseError` to attach to the response. The `RemoteEndpoint` will handle the exception and send a response message with the attached error object.

For example:
```

documentation/jsonrpc.md
```diff
@@ -29,11 +29,28 @@ You always work with two `Endpoints`. Usually one of the endpoints, a `RemoteEnd
...
+```java
+   @Override
+   public CompletableFuture<Object> shutdown() {
+      if (this.state == "not initialized") {
```
pisv, line 45:
```suggestion
      if (!isInitialized()) {
```

documentation/jsonrpc.md
```diff
@@ -29,11 +29,28 @@ You always work with two `Endpoints`. Usually one of the endpoints, a `RemoteEnd
...
+         ResponseError error = new ResponseError(ResponseErrorCode.serverNotInitialized, "Server was not initialized", null);
+         throw new ResponseErrorException(error);
+      }
+      this.state = "shutdown";
```
pisv, line 49:
```suggestion
      doShutdown();
```

pisv (commented):
Thanks for the proposal! It looks good, but I think we need to be a bit more careful with the wording. I've tried to come up with some suggestions.

fredg1:
Seemed good.

documentation/jsonrpc.md
```diff
@@ -29,11 +29,30 @@ You always work with two `Endpoints`. Usually one of the endpoints, a `RemoteEnd
...
+For example:
+```java
+   @Override
+   public CompletableFuture<Object> shutdown() {
+      if (!isInitialized()) {
+         ResponseError error = new ResponseError(ResponseErrorCode.serverNotInitialized, "Server was not initialized", null);
+         throw new ResponseErrorException(error);
+      }
+      doShutdown();
+      return null;
```
pisv, line 52:
```suggestion
      return doShutdown();
```

pisv (commented):
One more change. I don't think `null` is a valid return value for a request processing method.

documentation/jsonrpc.md
fredg1: https://microsoft.github.io/language-server-protocol/specifications/specification-current/#shutdown

`Shutdown` specifically needs to return `null`

fredg1:
> 
> 
> One more change. I don't think `null` is a valid return value for a request processing method.

Right, I forgot that all return values have to be wrapped in `CompletableFuture.completedFuture(<value>)`

pisv:
LGTM. But let us wait a bit and see if there are more comments.

@eclipse-lsp4j-bot ok to test

pisv (approved):
Since there have been no more comments, I assume we are good to go.

PR #563, opened 2021-07-23
Eclipse Feature contains incorrect license

mbooth101:
The lsp4j-feature still contains EPL v1.0 even though lsp4j is
now licensed under EPL v2.0. I noticed this because it is the
license from the feature that is shown to the user when they
are installing it into their IDE -- it was clearly incorrect.

This change makes the lsp4j-feature take advantage of the CBI
shared license functionality to generate and populate the license
information automatically at build time. The exact license to
use is specified in the target platform as the current latest
version of EPL v2.0.

Signed-off-by: Mat Booth <mat.booth@gmail.com>


pisv:
@eclipse-lsp4j-bot ok to test

pisv:
Thanks Mat for spotting the issue and providing a high-quality patch!

My only concern, not related to the patch itself, would be that LSP4J is dual-licensed under EPL 2.0 and EDL 1.0. Perhaps this is a special case that would preclude the project from using the CBI shared license. Other than that, I'm totally fine with your patch, which works as expected.

Let us wait until a project lead finds time to review.

releng/releng-target/lsp4j.target.target
```diff
@@ -13,5 +13,9 @@
 			<unit id="javax.websocket" version="0.0.0"/>
 			<unit id="javax.websocket.source" version="0.0.0"/>
 		</location>
+		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="false" type="InstallableUnit">
+			<repository location="http://download.eclipse.org/cbi/updates/license/"/>
+			<unit id="org.eclipse.license.feature.group" version="2.0.2.v20181016-2210"/>
```
jonahgraham, line 18: Minor change needed - this should be 0.0.0 version so that the latest version of the license is automatically picked up.

e.g: https://github.com/eclipse-cdt/cdt/blob/79b1f157dd9f51a535f23f7930ae3518feb526e3/releng/org.eclipse.cdt.target/cdt.target#L6

jonahgraham:
> Let us wait until a project lead finds time to review.

Hopefully @spoenemann just knows the answer on this - if not I will do some research to find out.

releng/releng-target/lsp4j.target.target
pisv: I am sorry, but would not that mean, in theory at least, that a new EPL version like 2.1.x or even 3.x might be inadvertently picked up? Would not it be safer to err on the side of specifying the exact version?

jonahgraham: The `0.0.0` is just standard practice across multiple Eclipse projects, it ensures that latest version is picked up. It does mean that it will become EPL3 automatically, but almost all project do it this way. I assume that it is to make sure we get the latest 2.0.x version. If target files allowed ranges (like [tpd](https://github.com/eclipse-cbi/targetplatform-dsl)) then the problem would be better solved. 

Of course, the EPL 2 license feature hasn't been updated in ~3 years now, so probably not a big issue.

BTW https://bugs.eclipse.org/bugs/show_bug.cgi?id=566576 was supposed to have documented this, but I guess it never happened.

pisv: I mentioned it only because in Handly we used the same approach as suggested by Mat to err on the side of caution with the legal stuff: https://git.eclipse.org/c/handly/org.eclipse.handly.git/tree/targets/base/base.target#n23. The downside is that the license version has to manually updated when a new revision becomes available and there might be no notification about it. The upside is that we don't run the risk of inadvertently picking up an inappropriate license.

Thanks for the reference to the bug. Although some links are broken, the last sentence is "If there is a new version of the license we update the version of the license feature we include in the target platform", which seems to indicate that the version is assumed to be manually maintained.

Having said that, I understand that specifying the `0.0.0` is a common practice across multiple Eclipse projects.

jonahgraham: It was a minor thing - the fact that other do it differently (like handly) means that we don't need to change this. Resolving.

mbooth101:
> > Let us wait until a project lead finds time to review.
> 
> Hopefully @spoenemann just knows the answer on this - if not I will do some research to find out.

FWIW I think my patch is "no worse" in that respect. The feature previously included a copy of EPL v1.0 only.

mbooth101:
*However* there's nothing preventing inclusion of additional licenses -- see the additional commit I made to this branch 9092680

pisv:
Thanks Mat for the clarification!

LGTM. It can be seen that the current patch produces the required legal artefacts for the feature as described in https://www.eclipse.org/projects/handbook/#legaldoc-plugins-features.

My concern is withdrawn. It turns out that it was based on a misunderstanding of the legal stuff for features. It is the [SUA](https://www.eclipse.org/projects/handbook/#software-user-agreement), not the actual license text, that is shown to the user when installing a feature. So, dual-licensing should indeed present no special case and the CBI shared license can still be used, as demonstrated by the current patch.

Unrelated to this patch, it seems that the [LICENSE](https://github.com/eclipse/lsp4j/blob/master/LICENSE) file should be amended to include the text of the EDL 1.0 license, as described in https://www.eclipse.org/projects/handbook/#legaldoc-license.

pisv:
> Unrelated to this patch, it seems that the [LICENSE](https://github.com/eclipse/lsp4j/blob/master/LICENSE) file should be amended to include the text of the EDL 1.0 license, as described in https://www.eclipse.org/projects/handbook/#legaldoc-license.

I have created PR #564 to address that.

spoenemann:
The project license is EPL **or** EDL (`EPL-2.0 OR BSD-3-Clause` in SPDX), so I think it's fine to mention only EPL in the context of an Eclipse feature installation.

pisv (approved):
Since the project leads have reviewed the patch and and no objections have been raised, I assume we are good to go. Thanks Mat for the great work!

PR #572, opened 2021-10-15
Update to Gradle 6

cdietrich:
update to gradle 6.9.2
replace osgi with bnd pluin
update to latest xtend gradle plugin
use maven-publish instead of maven plugin


cdietrich:
anyone have an idea why the PR builds are not working whist the branch build is? something broken how patches are applied to main and tested?

org.eclipse.lsp4j/build.gradle
```diff
@@ -17,17 +17,15 @@ sourceCompatibility = '1.8'
 targetCompatibility = '1.8'
 
 dependencies {
-	compile project(":org.eclipse.lsp4j.generator")
-	compile project(":org.eclipse.lsp4j.jsonrpc")
-	testCompile "junit:junit:$versions.junit"
+	api project(":org.eclipse.lsp4j.generator")
```
nixel2007, line 20: Are you sure you want to expose `generator` dependency to core lsp4j users?

cdietrich: did not want to change anything in first step.
https://search.maven.org/artifact/org.eclipse.lsp4j/org.eclipse.lsp4j/0.12.0/jar
dont know how and if it would work without

nixel2007: ah. generator is exposed already, sorry.
my point is about visibility of dependencies. If you migrate from `java` plugin to `java-library` you get new `api` configuration. in `java-library` consumer of your library can't see transitive dependencies from `implementation` configuration, but `api` dependencies are visible.
So *mayby* you don't want to just change `compile` dependencies to `api`, but split them: some of them to `api` configuration and some to `implementation` configuration.

cdietrich: yes but would prefer to do this in a later step (and by a person with better gradle knowledge) i also dont know if there is people already abusing it.

cdietrich:
@jonahgraham no as you can see Jenkins PR Build  is red but continuous-integration/jenkins/branch  is green

releng/build.Jenkinsfile
```diff
@@ -37,7 +37,7 @@ pipeline {
                 -PignoreTestFailures=true \
                 --refresh-dependencies \
                 --continue \
-                clean build signJar createLocalMavenRepo \
+                clean build signJar publish \
```
jonahgraham, line 40: Reason PR build is failing is that PR build does not use Jenkinsfile - either you need to make this same change to https://ci.eclipse.org/lsp4j/job/lsp4j-github-pullrequests/configure - or preferably change https://ci.eclipse.org/lsp4j/job/lsp4j-github-pullrequests to be Jenkinsfile based.

cdietrich: of course this will not work because different gradle versions

jonahgraham: The change to job can be done when this gets merged.

gradle/versions.gradle
```diff
@@ -20,5 +20,5 @@ ext.versions = [
 	'gson_orbit': '2.8.9.v20220111-1409',
 	'websocket_jakarta': '2.0.0',
```
cdietrich, line 21: should we bump xtend_lib and websocket_jakarta versions too?

jonahgraham: :+1: - the websocket jakarta needs a newer version in Orbit though for that to work???

ice1000:
Gradle 7.5.1 is released, are we going to update to Gradle 7?

cdietrich:
Unwound like to do it in two steps but first we need to do this one and I have no clue how to verify it

ice1000:
The ci fails with

```
04:36:01 FAILURE: Build failed with an exception.
 04:36:01
 04:36:01 * What went wrong: 
04:36:01 Task 'createLocalMavenRepo' not found in root project 'lsp4j'.
 04:36:01 
04:36:01 * Try: 
04:36:01 Run gradlew tasks to get a list of available tasks. Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
 04:36:01 
04:36:01 * Get more help at https://help.gradle.org 
04:36:01 
04:36:01 BUILD FAILED in 17s
```

cdietrich:
The problem is that the ci executes the old build with the new changes

cdietrich:
how shall we continue here

jonahgraham:
> how shall we continue here

Technical question or otherwise.

If it is ready and acceptable we should merge it and update the PR job so future checks don't fail https://github.com/eclipse/lsp4j/pull/572#discussion_r784045066

cdietrich:
I am still not sure if everything is still working as intended. Have no clue eg how to test if jars and poms are comparable

jonahgraham:
I'll try to review that - i.e. by manually looking at the before and after generated jars/poms.

cdietrich:
will give a look if some of the diffs can be addressed by configuring the plugins differently

cdietrich:
i can see Import-Package org.eclipse.lsp4j.jsonrpc.messages;version=
 "[0.16,1)" in the new variant too.

cdietrich:
ok, can you formally approve, then we can merge and i can test with snapshot consuming in Xtext

jonahgraham (approved):

PR #574, opened 2021-11-19
follow orbit and update gson to 2.8.8

cdietrich:
follow orbit and update gson to 2.8.8


cdietrich:
will update the PR to 2.8.9 (needs a while until orbit m2 is available)

PR #581, opened 2021-12-09
Improve descriptions and fixed typo

KamasamaK:
Improved some descriptions and fixed a typo in `TraceValue`.

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/TraceValue.java
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/TraceValue.java
@@ -23,7 +23,7 @@ public class TraceValue {
 
 	public static final String Off = "off";
 
-	public static final String Message = "message";
+	public static final String Message = "messages";
 
 	public static final String Verbose = "verbose";
 }
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/TraceValue.java
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/TraceValue.java
@@ -23,7 +23,13 @@ public class TraceValue {
 
 	public static final String Off = "off";
 
-	public static final String Message = "messages";
+	/**
+	 * @deprecated Use {@link #Messages} instead. This was a typo in the spec.
+	 */
+	@Deprecated
+	public static final String Message = "message";
+
+	public static final String Messages = "messages";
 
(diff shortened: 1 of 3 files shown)


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/TraceValue.java
```diff
@@ -23,7 +23,7 @@ private TraceValue() {
 
 	public static final String Off = "off";
 
-	public static final String Message = "message";
+	public static final String Message = "messages";
```
jonahgraham, line 26: Should this field be changed to "Messages" too for consistency with the surrounding ones?

jonahgraham (commented):
LGTM (+ a question, I don't mind answer either way) - thanks for the fixups.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/TraceValue.java
KamasamaK: Probably. I was just trying to avoid API breaking.

pisv: Should not we then deprecate the old constant and introduce a new one instead? Or, assuming no one has actually used it (given that no issue has been reported about it so far), could not we just go ahead and change both the constant name and value (with a corresponding note about the breaking change)? Anyway, changing the published constant value is already a breaking change, formally.

KamasamaK: Since it was actually a typo in the spec, I will go with deprecation.

pisv (approved):
Although it is not entirely clear to me what exactly 'when `falsy`' means in the spec, I think that 'when omitted or empty' is a good approximation in this context ;-) Thanks!

KamasamaK:
@pisv Since the spec is written in TypeScript, it just refers to [`falsy` in a JavaScript context](https://developer.mozilla.org/en-US/docs/Glossary/Falsy). I felt it was appropriate to translate for our use.

pisv:
> @pisv Since the spec is written in TypeScript, it just refers to [`falsy` in a JavaScript context](https://developer.mozilla.org/en-US/docs/Glossary/Falsy). I felt it was appropriate to translate for our use.

@KamasamaK Thanks for the link! I now see that it is defined precisely, and the translation is quite exact!

Issue #547, opened 2021-04-20
Migrate to Java 11


jonahgraham:
Lets migrate to Java 11. 

Any objections?

cdietrich:
is the current release 3.16 complete or is there still something missing?

rgrunber:
JDT-LS  made Java 11 a requirement a while ago (I believe due to Platform switching), and you can see much of the reaction to that at https://marketplace.visualstudio.com/items?itemName=redhat.java&ssr=false#review-details .

Not saying you should continue to support it despite difficulties, but if the work required to continue supporting is minimal it would reduce such complaints.

mickaelistria:
>  if the work required to continue supporting is minimal it would reduce such complaints.

I don't imagine end-users would typically complain about LSP4J version, it's an internal detail for them. So it's more about whether some LSP4J-based clients or server do care enough about Java 8 to make it still worth being supported.
I don't think any form of complaints is worth keeping support for Java 8; unless those complaints are backed by contributions to match them.
Java 8 support already did cost me about 20minutes + some CI cycles + probably 15 to 20 minutes to Jonah to figure out my mistake, and maybe more unnoticed more. I am personally not caring about Java 8 so this was pure loss to me, so if people care about Java 8 for LSP4J enough, I'd like their address to send them the bill of extra time Java 8 has cost today ;)

jonahgraham:
> is the current release 3.16 complete or is there still something missing?

If I am reading between the lines correctly - do you recommend calling 3.16 done done before moving to Java 11. That is fine to me.

> i am not sure if we at Xtext already want to ditch Java 8 too.

If Xtext moved that would signal 100% that LSP4J should move. Xtext has much wider user base AFAICT, many of them using it for years before LSP4J's existence.

> maybe source 11 target 8 could be an option?

I don't think so - what spurred this today was using new library methods (specifically List.of today).

jonahgraham:
> JDT-LS made Java 11 a requirement a while ago (I believe due to Platform switching), and you can see much of the reaction to that at https://marketplace.visualstudio.com/items?itemName=redhat.java&ssr=false#review-details .

That was a depressing read. Although it looked a lot like the hate when changing from Java 7 to 8. Nice to see some positive comments in there too.

> Not saying you should continue to support it despite difficulties, but if the work required to continue supporting is minimal it would reduce such complaints.

[...]

PR #591, opened 2022-01-13
update to tycho 1.7

cdietrich:
update to tycho 1.7

### DIFF
--- a/releng/p2/pom.xml
+++ b/releng/p2/pom.xml
@@ -13,7 +13,7 @@
 	</parent>
 
 	<properties>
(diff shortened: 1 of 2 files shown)


cdietrich:
no we would need to switch to java 11 then

jonahgraham:
> no we would need to switch to java 11 then

:-) We are supposed to do that for 0.13.0 #547

cdietrich:
yes, but i think we should do 1 thing at a time. i am not sure if i will find any time to work on it

PR #597, opened 2022-03-23
Add WorkspaceSymbol resolve

KamasamaK:
- Add `resolveWorkspaceSymbol`
- Change return type for `symbol` (Breaking change)
- Deprecate `SymbolInformation`


KamasamaK:
I've always just copy-pasted for these new files initially, including the copyright notice. I changed it to have the correct year and me.

PR #599, opened 2022-03-23
Update Type Hierarchy

KamasamaK:
Updated Type Hierarchy based on latest spec. This will replace the non-standard implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,13 @@ Breaking API changes:
  * Return type of `workspace/symbol` changed from `List<? extends SymbolInformation>` to
    `Either<List<? extends SymbolInformation>, List<? extends WorkspaceSymbol>>`
 
+Breaking Beta API changes:
+
+ * Significant updates were made to the `TypeHierarchy` features to replace the non-standard implementation
+
 Deprecated API changes:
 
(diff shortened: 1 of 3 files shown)


jonahgraham (commented):
LGTM. I have done a once over on this, but not a detailed review. It is great to keep the code up to date with the 3.17 spec as it develops.

I think a line in the changelog here would be good too as (at least) https://github.com/eclipse/eclipse.jdt.ls uses typeHierarchy and it will help them resolve the compilation errors when they update.

KamasamaK:
I will update the change log in a separate PR with all of the notable changes collected.

Would this be considered a "Breaking API change"? I would assume anything annotated with `@Beta` would be exempt from that.

jonahgraham:
> I will update the change log in a separate PR with all of the notable changes collected.
> 
> Would this be considered a "Breaking API change"? I would assume anything annotated with `@Beta` would be exempt from that.

Certainly adding it to there is a courtesy. Since I know that consumers of lsp4j use that api I thought it would be easier for them to see why there code does not compile. It can go under a separate heading if you prefer, or even omitted.

KamasamaK:
FYI @mickaelistria and @angelozerr since you work on projects that use LSP4J as a client and server, respectively.

PR #600, opened 2022-03-24
Add inlay hint support

KamasamaK:
Adds inlay hint support which is currently proposed for LSP 3.17


nixel2007:
First naive implementation for BSL Language Server. Thanks a lot, @KamasamaK!
image

mickaelistria:
That's a feature we'd love to integrate ASAP in LSP4E, and the current patch seems correct. We'd appreciate if it can be merged and then released soon-ish ;)

angelozerr:
+1, JDT LS and Qute LS require this PR. Today they use their own InlayHint Java class, but when LSP4J wil be released with this InlayHint support we could remove those cutsom class.

mickaelistria:
Thanks @KamasamaK !

PR #613, opened 2022-05-05
Smaller 3.17 changes and various others

KamasamaK:
### LSP 3.17 changes

- Add position encoding support
- Add relative pattern support
- Add `CompletionItem.textEditText`
- Add `FoldingRange.collapsedText`
- Add Folding Range kind capabilities

### Breaking changes

- Removed `RenameOptions.id` as it was never specified for `StaticRegistrationOptions`
- Type of `FileSystemWatcher.globPattern` changed from `String` to `Either<String, RelativePattern>`
- Removed `WorkspaceEdit.resourceChanges`, `WorkspaceEditCapabilities.resourceChanges`, and `ResourceChange` as they have been deprecated for numerous versions and are not specified

### Other changes

- Add more constructors
- Improve some docblocks
- Move some classes with only static strings into `Protocol.xtend`


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -1403,6 +1401,54 @@ class DiagnosticsTagSupport {
...
+@Beta
+@JsonRpcData
+class FoldingRangeKindSupportCapabilities {
+	/**
+	 * The folding range kind values the client supports. When this
+	 * property exists the client also guarantees that it will
+	 * handle values outside its set gracefully and falls back
+	 * to a default value when unknown.
+	 * <p>
+	 * See {@link FoldingRangeKind} for some predefined folding range kinds.
```
cdietrich, line 1418: dead link?

cdietrich: i see was moved to xtend

KamasamaK: Yes. I'm planning to move some more over in a separate PR. I'm not sure why they're separated.

PR #615, opened 2022-05-11
Various minor changes

KamasamaK:
### Breaking changes

- Removed `RenameOptions.id` as it was already deprecated and never specified for `StaticRegistrationOptions`
- Removed `SemanticTokenTypes.Member` as it was already deprecated and is not specified
- Removed `TraceValue.Message` as it was already deprecated and is not specified
- Changed `TraceValue` to be `final` matching similar classes
- Removed duplicate `ResponseErrorCode` as it has been deprecated for several versions
- Removed `ResponseErrorCode.serverErrorStart` and `ResponseErrorCode.serverErrorEnd` as they were already deprecated and just boundaries not actual error codes

### Other changes

- Improve many docblocks
- Move many classes with only static members into `Protocol.xtend`
- Improve JSON-RPC Markdown document

PR #617, opened 2022-05-13
 fix some javadocs

cdietrich:
fix some javadocs


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/ResponseErrorCode.java
```diff
@@ -1,46 +0,0 @@
-/******************************************************************************
```
KamasamaK, line 1: I'm already removing this in #615.

KamasamaK (changes requested):

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/ResponseErrorCode.java
cdietrich: somehow my search in your pr did not find it :(

KamasamaK (approved):

PR #618, opened 2022-05-13
Added missing Type Adaper for RestartArguments.arguments

cdietrich:
Added missing Type Adaper for RestartArguments.arguments

### DIFF
--- a/org.eclipse.lsp4j.debug/src/test/java/org/eclipse/lsp4j/debug/test/RestartArgumentsTypeAdapterFactoryTest.java
+++ b/org.eclipse.lsp4j.debug/src/test/java/org/eclipse/lsp4j/debug/test/RestartArgumentsTypeAdapterFactoryTest.java
@@ -1,0 +1,38 @@
+/******************************************************************************
+ * Copyright (c) 2022 itemis AG and others.
+ * 
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0,
+ * or the Eclipse Distribution License v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ * 
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ ******************************************************************************/
+package org.eclipse.lsp4j.debug.test;
+
+import static org.junit.Assert.assertTrue;
+
+import org.eclipse.lsp4j.debug.RestartArguments;
+import org.junit.Test;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+
+public class RestartArgumentsTypeAdapterFactoryTest {
+
+	@Test
+	public void test() {
+		GsonBuilder builder = new GsonBuilder();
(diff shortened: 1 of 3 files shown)


org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/adapters/RestartArgumentsTypeAdapterFactory.java
```diff
@@ -0,0 +1,45 @@
...
+
+public class RestartArgumentsTypeAdapterFactory implements TypeAdapterFactory {
+
+	private static final TypeToken<Either<LaunchRequestArguments, AttachRequestArguments>> ELEMENT_TYPE = new TypeToken<Either<LaunchRequestArguments, AttachRequestArguments>>() {
+	};
+
+	@SuppressWarnings("unchecked")
+	@Override
+	public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
+		if (!ELEMENT_TYPE.equals(type))
```
cdietrich, line 36: is this correct. without it will stackoverflow

org.eclipse.lsp4j.debug/src/test/java/org/eclipse/lsp4j/debug/test/RestartArgumentsTypeAdapterFactoryTest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.eclipse.lsp4j.jsonrpc.messages.Either;
+import org.junit.Test;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+
+public class RestartArgumentsTypeAdapterFactoryTest {
+
+	@Test
+	public void test() {
```
cdietrich, line 28: is there a better way to test such things?

pisv:
LGTM, with just a few marginal notes:

* Names of existing type adapter factories in `org.eclipse.lsp4j.adapters` end with `Adapter` rather than `AdapterFactory`. So it seems more appropriate to name this class `RestartArgumentsTypeAdapter` to match the already established naming convention, although the name `RestartArgumentsTypeAdapterFactory` might be more correct from the technical point of view.

* The test case looks fine to me. Note that this is the first test case for type adapter factories in LSP4J, so there are currently no established convention on how to test such things.

* I think the check `ELEMENT_TYPE.equals(type)` is correct and is necessary if the type adapter factory is registered globally with `GsonBuilder` (as done in the test case). 

  Existing type adapter factories, which miss this check, don't have the mentioned problem with stackoverlow only because they are never registered globally, but are targeted precisely via annotations.

  It would probably be good to also add this check to existing type adapter factories to make the code more tolerant to usage in different contexts.

cdietrich:
@pisv i adapted the test to test the annotated type.
now if would work without the type check
should i remove it?

pisv:
Perhaps the test case could be slightly improved by using `assertEquals` and comparing the deserialized object with the expected value.

cdietrich:
done

cdietrich:
created https://github.com/eclipse/lsp4j/issues/623 as follow up

PR #625, opened 2022-05-17
Add japicmp script

jonahgraham:
This script wraps https://siom79.github.io/japicmp/CliTool.html
to create a set of HTML reports for each bundle showing
the API changes since a baseline.

Here are the current reports: [apicmp-report.zip](https://github.com/eclipse/lsp4j/files/8711561/apicmp-report.zip)

### DIFF
--- a/releng/build.Jenkinsfile
+++ b/releng/build.Jenkinsfile
@@ -54,6 +54,13 @@ pipeline {
         }
       }
     }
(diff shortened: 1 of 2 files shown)


jonahgraham:
Abandoning in favour of a branch in the main repo so that I can test releng before merging

PR #630, opened 2022-05-25
Late updates for LSP 3.17

KamasamaK:
These are the changes to the LSP 3.17 specification that were added after it was published as done.

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/PrepareSupportDefaultBehavior.java
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/PrepareSupportDefaultBehavior.java
@@ -21,7 +21,7 @@ public enum PrepareSupportDefaultBehavior {
 	
 	/**
 	 * The client's default behavior is to select the identifier
-	 * according the to language's syntax rule.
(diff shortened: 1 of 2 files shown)


jonahgraham:
We can spin up a 0.14.0 release if needed in short order, please let me know with a comment on #628

cdietrich:
i assume a 0.13.1 would be better for p2 consumers
as maven consumers a rebuild would be needed anway

jonahgraham:
> i assume a 0.13.1 would be better for p2 consumers as maven consumers a rebuild would be needed anway

I advise against it because there are API changes in this commit - 0.13.0 and 0.13.1 are not interchangeable due to new fields and the changed super type hierarchy. I could be convinced otherwise though if you think it is unlikely enough for anyone to hit that problem in practice, especially as we are still officially incubating.

cdietrich:
Yes. The question is when it would be available. I planned to do a Xtext release on Monday

jonahgraham:
@cdietrich I can release very quick once it is merged as I just went through the cycle of that a few days ago. Can you or @KamasamaK  or someone else approve and merge?

jonahgraham (approved):

cdietrich (approved):

PR #635, opened 2022-06-09
Another late addition for LSP 3.17

KamasamaK:
Added `WorkspaceSymbol.data`

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -6800,6 +6800,13 @@ class WorkspaceSymbol {
 	 */
 	String containerName
 
+	/**
+	 * A data entry field that is preserved on a workspace symbol between a
+	 * workspace symbol request and a workspace symbol resolve request.
+	 */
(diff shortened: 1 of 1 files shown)


jonahgraham:
Is there an urgent need to roll this out? Or can it wait? It is frustrating that a "final" spec keeps changing. I can see the new field was added in the typescipt code 7 months ago: https://github.com/microsoft/vscode-languageserver-node/blame/1fe84c5b656d9db07f812cb573d2c0d82f9d8366/client/src/common/protocolWorkspaceSymbol.ts#L11

Either way I think this should be updated:

https://github.com/eclipse/lsp4j/blob/95c0ed6c90192e2a8ab12b1ccb2e142eef476b15/README.md?plain=1#L38-L41

KamasamaK:
I agree. These should have probably been 3.17.1 or something like that. I don't feel it's urgent, but I didn't think the previous additions were either.

Either way, I will try to look into Notebook support soon so that the next release is more worthwhile.

jonahgraham:
> I agree. These should have probably been 3.17.1 or something like that. I don't feel it's urgent, but I didn't think the previous additions were either.

For the previous ones we had just released and I wanted to get (What I had hoped) was the real final 3.17.0 into the 2022-06 SimRel release.

> 
> Either way, I will try to look into Notebook support soon so that the next release is more worthwhile.

Great.

Issue #632, opened 2022-06-02
Tomcat TEXT_FULL_WRITING state exception while multiple messages sent from server to client


ivy-lli:
Hi 

We use [GLSP](https://github.com/eclipse-glsp/glsp) to display a diagram. GLSP itself uses LSP (LSP4J) as communication layer between server and client. 
As our application already has a Tomcat web server we wanted to reuse (same port etc.) this one instead of provide something new like Jetty.

But now we run into many exceptions in our log, that a message couldn't be sent to the client:
```log
[WARN ][org.eclipse.lsp4j.jsonrpc.RemoteEndpoint][IvyGlspActionExecutor-1]{}
Failed to send notification message.
java.lang.IllegalStateException: The remote endpoint was in state [TEXT_FULL_WRITING] which is an invalid state for called method
	at org.apache.tomcat.websocket.WsRemoteEndpointImplBase$StateMachine.checkState(WsRemoteEndpointImplBase.java:1274)
	at org.apache.tomcat.websocket.WsRemoteEndpointImplBase$StateMachine.textStart(WsRemoteEndpointImplBase.java:1236)
	at org.apache.tomcat.websocket.WsRemoteEndpointImplBase.sendStringByCompletion(WsRemoteEndpointImplBase.java:213)
	at org.apache.tomcat.websocket.WsRemoteEndpointImplBase.sendStringByFuture(WsRemoteEndpointImplBase.java:201)
	at org.apache.tomcat.websocket.WsRemoteEndpointAsync.sendText(WsRemoteEndpointAsync.java:53)
	at org.eclipse.lsp4j.websocket.WebSocketMessageConsumer.sendMessage(WebSocketMessageConsumer.java:57)
	at org.eclipse.lsp4j.websocket.WebSocketMessageConsumer.consume(WebSocketMessageConsumer.java:47)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.notify(RemoteEndpoint.java:126)
	at org.eclipse.lsp4j.jsonrpc.services.EndpointProxy.invoke(EndpointProxy.java:88)
	at com.sun.proxy.$Proxy102.process(Unknown Source)
	at org.eclipse.glsp.server.actions.ClientActionHandler.send(ClientActionHandler.java:58)
	at org.eclipse.glsp.server.actions.ClientActionHandler.execute(ClientActionHandler.java:52)
[...]

PR #638, opened 2022-06-21
#632: send messages with a queue each after each

ivy-lli:
Fix for: https://github.com/eclipse/lsp4j/issues/632

I tried to implement a queue for sending messages over WebSocket.
I used a SendHandler to define a callback, where the next message is get from the queue and sent to the client.

With this change messages are sent each after another. So also the new test I've added work. Without the queue this test fails with a random order of the numbers.

Also if I now run my tests against a such patched LSP4J lib, I don't receive the `TEXT_FULL_WRITING` exception any longer from Tomcat, and the tests are much more stable.

What do you think? Do you have any concerns or improvements?

### DIFF
--- a/org.eclipse.lsp4j.websocket.jakarta/src/main/java/org/eclipse/lsp4j/websocket/jakarta/WebSocketMessageConsumer.java
+++ b/org.eclipse.lsp4j.websocket.jakarta/src/main/java/org/eclipse/lsp4j/websocket/jakarta/WebSocketMessageConsumer.java
@@ -12,9 +12,13 @@
 package org.eclipse.lsp4j.websocket.jakarta;
 
 import java.io.IOException;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import jakarta.websocket.SendHandler;
+import jakarta.websocket.SendResult;
 import jakarta.websocket.Session;
 
 import org.eclipse.lsp4j.jsonrpc.JsonRpcException;
@@ -31,6 +35,8 @@ public class WebSocketMessageConsumer implements MessageConsumer {
 	
 	private final Session session;
 	private final MessageJsonHandler jsonHandler;
+	private final ConcurrentLinkedQueue<String> messageQueue = new ConcurrentLinkedQueue<>();
+	private final WebSocketSendHandler handler = new WebSocketSendHandler();
 	
 	public WebSocketMessageConsumer(Session session, MessageJsonHandler jsonHandler) {
 		this.session = session;
@@ -55,7 +61,8 @@ public class WebSocketMessageConsumer implements MessageConsumer {
 		if (session.isOpen()) {
 			int length = message.length();
 			if (length <= session.getMaxTextMessageBufferSize()) {
-				session.getAsyncRemote().sendText(message);
+				messageQueue.add(message);
+				handler.handleNextMessage();
 			} else {
 				int currentOffset = 0;
 				while (currentOffset < length) {
@@ -66,6 +73,22 @@ public class WebSocketMessageConsumer implements MessageConsumer {
 			}
 		} else {
 			LOG.log(Level.INFO, "Ignoring message due to closed session: {0}", message);
+		}
+	}
+	
+	private class WebSocketSendHandler implements SendHandler {
+		private final AtomicBoolean isSending = new AtomicBoolean();
+		
+		@Override
+		public void onResult(SendResult result) {
+			isSending.set(false);
+			handleNextMessage();
+		}
+		
+		void handleNextMessage() {
+			if (session.isOpen() && !messageQueue.isEmpty() && isSending.compareAndSet(false, true)) {
+				session.getAsyncRemote().sendText(messageQueue.poll(), this);
+			}
 		}
 	}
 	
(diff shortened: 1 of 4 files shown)


cdietrich:
ci currently does not work cause of https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1443

pisv:
@eclipse-lsp4j-bot ok to test

pisv (commented):
@ivy-lli Thanks for the patch!

I'm not a web socket expert in any way, but have taken a look at your patch and will try to provide some general feedback.

The approach looks good to me from the concurrency point of view. The code also looks good, besides some minor nitpicking regarding visibility modifiers.

I have verified locally that the unit test you provided fails occasionally without the change, and succeeds with the change. On my machine, I had to increase the number of iterations to 20 to make the test fail "more reliably" without the change (FWIW, I noticed that occasionally it would fail because of a timeout in this case).

Please note that there is also `org.eclipse.lsp4j.websocket.jakarta` project, where a similar change needs to be applied as well.

Thanks again for reporting the bug and providing a high-quality patch.

ivy-lli:
Hi @pisv 

Thanks for the feedback!
I'm aware of the jakarta project and that it's need the same changes, but I wanted to make this Draft PR first to get a general feedback, before make the changes there.

I will adjust my PR with the changes you supposed.

pisv:
> I'm aware of the jakarta project and that it's need the same changes, but I wanted to make this Draft PR first to get a general feedback, before make the changes there.

Makes total sense!

> I will adjust my PR with the changes you supposed.

Thank you! The patch in its current form looks good to me, but let us wait a bit to give other committers opportunity to review, ok?

jonahgraham:
This code change looks good to me. My concern is that the new test does not fail when I try to run it on the original code - is there some additional setup you do to make it fail? I tried to make test fail with Java 8 and Java 11 on Linux.

pisv:
FWIW, on my machine the new test is flaky (fails about every third time) against the original code. When I increase the number of iterations from 10 to 20, like

```java
	@Test
	public void testManyConcurrentNotifications() throws Exception {
		for (int i = 0; i < 20; i ++) {
			client.server.notify(String.valueOf(i % 10));
		}
		await(() -> server.result.length() == 20);
		Assert.assertEquals("01234567890123456789", server.result);
	}
```

the test fails repeatedly. In both cases the test succeeds constantly with the patch applied locally.

So, it looks like 10 iterations might actually be not enough for reproducing the issue more or less reliably.

pisv:
I also noticed that sometimes the test failed with


```
java.lang.AssertionError: Timeout elapsed while waiting for condition.

	at org.junit.Assert.fail(Assert.java:88)
	at org.eclipse.lsp4j.websocket.test.MockConnectionTest.await(MockConnectionTest.java:100)
	at org.eclipse.lsp4j.websocket.test.MockConnectionTest.testManyConcurrentNotifications(MockConnectionTest.java:76)
```

when run on the original code (with both 10 and 20 iterations). I don't know whether this is to be expected or not.

ivy-lli:
I can increase the number of iterations if you like so.
I think it fails sometimes with this timeout because some of the messages are lost (e.g. if you print the incoming messages), but I'm not 100% sure. Because the other tests are always waiting for the message to arrive before sending the next one.

pisv:
> I can increase the number of iterations if you like so

Thanks @ivy-lli! Let us see if this helps @jonahgraham reproduce the issue locally. While we are experimenting, I have converted this PR to draft to remind us that there is another test in jakarta that may need to be updated as well,
so the PR is not yet ready to be merged.

> I think it fails sometimes with this timeout because some of the messages are lost

Yes, I think this might well be the reason, given the nature of the bug.

ivy-lli:
> I have converted this PR to draft to remind us that there is another test in jakarta that may need to be updated as well,
so the PR is not yet ready to be merged.

Sorry, I intended to also update the jakarta test, but I forgot to raise the interations there...

jonahgraham:
I changed the test locally so that I could read a bit better what is happening in the error log and to try to verify that the timeout was really due to lost messages:

```java	
	@Test
	public void testManyConcurrentNotifications() throws Exception {
		String expectedResult = "";
		for (char i = 'a'; i < 'z'; i ++) {
			String x = Character.toString(i);
			client.server.notify(x);
			expectedResult += x;
		}
		int expectedResultLenght = expectedResult.length();
		try {
			await(() -> server.result.length() == expectedResultLenght);		
		} catch (Error e) {
			// discard this error so that the nice error displays in the assertEquals
		}
		Assert.assertEquals(expectedResult, server.result);
	}
```

In the case that the await timedout I see *expected:&lt;abcdefghijklm[nopqrstuv]wxy&gt; but was:&lt;abcdefghijklm[potrquvs]wxy&gt;"* - the `n` is missing.

With the 10 count test I had a single failure once. I have run the 26 test a bunch of times and it fails consistently for me. I think this test is good enough. Interestingly the failures are almost always after the `j` test, i.e. the first 10 or so don't get lost and stay in order.

jonahgraham (approved):

pisv:
@jonahgraham I like your approach. Actually, I would vote for using your test, as it might provide better visibility into what is going on in case of regressions.

@ivy-lli WDYT?

ivy-lli:
@pisv, @jonahgraham 
Yes, a very good suggestion, I like it!

pisv (approved):
Thanks for the great work!

ivy-lli:
Thanks for the quick and professional feedback 😀

PR #642, opened 2022-06-27
Add Notebook support

### LINKED ISSUE #640: Notebook support
I'd love to start using 3.17s Notebook features. Is there a plan to start supporting them?

KamasamaK:
Fixes #640

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -2250,7 +2250,15 @@ class NotebookDocumentClientCapabilities {
 	/**
 	 * Capabilities specific to notebook document synchronization
 	 */
+	@NonNull
 	NotebookDocumentSyncClientCapabilities synchronization
+
+	new() {
+	}
+
+	new(@NonNull NotebookDocumentSyncClientCapabilities synchronization) {
+		this.synchronization = Preconditions.checkNotNull(synchronization, 'synchronization')
+	}
 }
 
 /**
@@ -10604,26 +10612,6 @@ class NotebookCell {
 	new(int kind, @NonNull String document) {
 		this.kind = kind
 		this.document = Preconditions.checkNotNull(document, 'document')
-	}
-}
-
-/**
- * A notebook cell kind.
- * <p>
- * Since 3.17.0
- */
-final class NotebookCellKind {
-	/**
-	 * A markup-cell is formatted source that is used for display.
-	 */
-	public static int Markup = 1
-
-	/**
-	 * A code-cell is source code.
-	 */
-	public static int Code = 2
-
-	private new() {
 	}
 }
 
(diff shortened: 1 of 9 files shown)


pisv (commented):
I have checked it against the spec. Great work, as always! Minor suggestions attached.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -10484,3 +10510,682 @@ class DiagnosticWorkspaceCapabilities {
...
+		this.document = Preconditions.checkNotNull(document, 'document')
+	}
+}
+
+/**
+ * A notebook cell kind.
+ * <p>
+ * Since 3.17.0
+ */
+final class NotebookCellKind {
```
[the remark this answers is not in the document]
pisv: Good points! Thanks for the clarification.

FWIW, the current approach will not fail on an unknown enumeration value - `null` is returned in this case (see `EnumTypeAdapter`), but it cannot preserve the value on round trips (the latter being a recommendation, not a requirement of the spec).

IMHO, it would be a more clear workflow to make it an enum for consistency sake now and discuss in a separate issue if a new pattern for enums needs to be established.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -10484,3 +10518,660 @@ class DiagnosticWorkspaceCapabilities {
...
+ * notebook cell or the cell's text document.
+ * <p>
+ * Since 3.17.0
+ */
+@JsonRpcData
+class NotebookCell {
+	/**
+	 * The cell's kind
+	 */
+	NotebookCellKind kind
```
pisv, line 10586: Looks like it should be `@NonNull`.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -10484,3 +10518,660 @@ class DiagnosticWorkspaceCapabilities {
...
+	/**
+	 * Additional execution summary information
+	 * if supported by the client.
+	 */
+	ExecutionSummary executionSummary
+
+	new() {
+	}
+
+	new(int kind, @NonNull String document) {
```
pisv, line 10610: Also, in constructor.

pisv (commented):
LGTM. Thanks!

Cyrik:
So I tried using the PR and I noticed that `org.eclipse.lsp4j.services.LanguageServer` has no `getNotebookService` method.  Is there some other way to plugin my own `NotebookService` implementation?

KamasamaK:
@Cyrik Sorry. This is my first time adding a new service, so I missed some things.

Cyrik:
No problem at all, thank you for the fast update! I didn't notice the missing bit either until I wanted to pass in my service 😄

pisv:
It looks like [this build](https://ci.eclipse.org/lsp4j/job/lsp4j-github-pullrequests/328/artifact/build/) includes the latest changes, does not it?

Cyrik:
ah yes, thank you. I didn't find that link and was trying to use the nightly or build it myself 😅 
The one you posted worked and after a bit of jvm version switching, I can also build my own. I've got capabilities to work from Clojure, going to try the actual sync messages tomorrow. Thank you.

PR #644, opened 2022-07-06
DAP v1.56

KamasamaK:
Implemented DAP version 1.56


org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/DebugProtocol.xtend
```diff
@@ -3689,6 +3686,38 @@ class StepInTarget {
 	 */
 	@NonNull
 	String label;
+	/**
+   * An optional line of the step in target.
```
szarnekow, line 3690: The indentation appears to be inconsistent here.

Issue #611, opened 2022-05-02
Allow more versions of gson?


frankbenoit:
Currently only exactly gson 2.8.9 is allowed.
Would it be possible to allow a wider range? 
if the semantic versioning is valid, even [2.8.9,3.0) should be ok.
This would simplify the coexistence with other bundles that have dependencies to gson.

jonahgraham:
Hi @frankbenoit, that sounds reasonable. Please can you provide a patch?

cdietrich:
https://github.com/eclipse/lsp4j/blob/32ad52691b2fcaef7011ffc6461641f1164dc1f2/gradle/versions.gradle#L19
isnt this one used for manifest too?

cdietrich:
and were there some breaking changes or do they do sem. versioning?

frankbenoit:
The minimum supported Java version changes from 6 to 7.
https://github.com/google/gson/releases/tag/gson-parent-2.9.0

nixel2007:
> and were there some breaking changes or do they do sem. versioning?

Google? semver and no breaking changes? Ha! They can break backward compatibility in patch release.

cdietrich:
we will go with [2.9.0,2.10) for now

PR #646, opened 2022-07-09
switch to orbit 2022-09 and widen version range for gson. See #611

cdietrich:
switch to orbit 2022-09 and widen version range for gson. See #611

unfortunately gson 2.9 is still not available in an orbit milestone. so i have to keep the 2.8.9 on p2 and the lower bound

### DIFF
--- a/releng/lsp4j-feature/feature.xml
+++ b/releng/lsp4j-feature/feature.xml
@@ -31,7 +31,7 @@
    </license>
 
    <requires>
-      <import plugin="com.google.gson" version="2.8.9" match="equivalent"/>
+      <import plugin="com.google.gson" version="2.9.0" match="equivalent"/>
       <import plugin="org.eclipse.xtend.lib" version="2.10.0" match="compatible"/>
    </requires>
 
--- a/releng/releng-target/lsp4j.target.target
+++ b/releng/releng-target/lsp4j.target.target
@@ -3,13 +3,19 @@
 <target name="org.eclipse.xtext.helios.target" sequenceNumber="0">
 	<locations>
 		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="false" type="InstallableUnit">
-			<repository location="https://download.eclipse.org/tools/orbit/downloads/2022-03"/>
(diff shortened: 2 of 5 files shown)


releng/releng-target/lsp4j.target.target
```diff
@@ -3,13 +3,16 @@
...
-			<repository location="https://download.eclipse.org/tools/orbit/downloads/2022-03"/>
+			<repository location="https://download.eclipse.org/tools/orbit/downloads/2022-09"/>
 			<unit id="com.google.gson" version="2.8.9.v20220111-1409"/>
 			<unit id="com.google.gson.source" version="2.8.9.v20220111-1409"/>
 			<unit id="com.google.guava" version="30.1.0.v20210127-2300"/>
 			<unit id="com.google.guava.source" version="30.1.0.v20210127-2300"/>
 			<unit id="jakarta.websocket" version="0.0.0"/>
 			<unit id="jakarta.websocket.source" version="0.0.0"/>
+		</location>
+		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="false" type="InstallableUnit">
```
cdietrich, line 14: https://github.com/eclipse/lsp4j/issues/647

releng/lsp4j-feature/feature.xml
```diff
@@ -31,7 +31,7 @@
    </license>
 
    <requires>
-      <import plugin="com.google.gson" version="2.8.9" match="equivalent"/>
+      <import plugin="com.google.gson" version="2.8.9" match="compatible"/>
```
cdietrich, line 34: i am not sure if we just should go to 2.9.0 as minimal version.

cdietrich: @jonahgraham what do you think?

jonahgraham: Lets make 2.9.0 min version.

cdietrich: done

jonahgraham (approved):
LGTM - once I do the Orbit S-build we can update the URL.

PR #649, opened 2022-07-15
Adjust inheritance tree for `SemanticTokensWithRegistrationOptions`

### LINKED ISSUE #573: Adjust inheritance tree for SemanticTokensWithRegistrationOptions
`SemanticTokensWithRegistrationOptions` currently extends `AbstractWorkDoneProgressOptions`. It should probably extend `AbstractTextDocumentRegistrationAndWorkDoneProgressOptions` instead. This would make it extend `TextDocumentRegistrationOptions`, from which other similar options are derived (and hence, would allow a greater opportunity for polymorphism).

Formally, this is a breaking API change, but it seems to be a relatively mild one in practice: because `AbstractWorkDoneProgressOptions` implements the interface `WorkDoneProgressOptions`, there seems to be little reason for LSP4J consumers to depend on the fact that `SemanticTokensWithRegistrationOptions` currently extends `AbstractWorkDoneProgressOptions` (as opposed to just implementing `WorkDoneProgressOptions`).

WDYT?

pisv:
Fixes #573

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@
 
 Fixed issues: <https://github.com/eclipse/lsp4j/milestone/22?closed=1>
 
+Breaking API changes:
+
+ * Class `SemanticTokensWithRegistrationOptions` now extends
+   `AbstractTextDocumentRegistrationAndWorkDoneProgressOptions`
+   instead of `AbstractWorkDoneProgressOptions`
+
 Nightly japicmp report: <https://download.eclipse.org/lsp4j/builds/main/japicmp-report/>
 
 ### v0.14.0 (May 2022)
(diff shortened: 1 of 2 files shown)


KamasamaK (approved):
LGTM.

I'm curious can you give an example of a usage where the API would break from this?

pisv:
Thanks for the review.

> I'm curious can you give an example of a usage where the API would break from this?

It would break if there were a cast to `AbstractWorkDoneProgress` in pre-existing client code, which seems unlikely in practice: https://github.com/eclipse/lsp4j/issues/573#issue-1054754518. But formally, this is a breaking change: https://wiki.eclipse.org/Evolving_Java-based_APIs_2#Evolving_API_Classes.

PR #650, opened 2022-07-17
Add support for client default behavior on renames

pisv:
See https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_prepareRename


pisv:
Note that this is the first time a type adapter for `Either3` is used in LSP4J. See `PrepareRenameResponseAdapter`. Some changes in `EitherTypeAdapter` were required to support it.

PR #663, opened 2022-09-11
DAP v1.59 updates

### LINKED ISSUE #662: Support new DAP `startDebugging` request
See https://github.com/microsoft/debug-adapter-protocol/issues/79 and https://github.com/microsoft/debug-adapter-protocol/pull/326

KamasamaK:
Updates DAP to v1.59.

- Update documentation
- Add `argsCanBeInterpretedByShell` property to `runInTerminal` request
- Add `startDebugging` request

Resolves #662


org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/adapters/StartDebuggingRequestArgumentsConfigurationTypeAdapter.java
```diff
@@ -0,0 +1,42 @@
...
+import com.google.gson.reflect.TypeToken;
+
+public class StartDebuggingRequestArgumentsConfigurationTypeAdapter implements TypeAdapterFactory {
+	private static final TypeToken<Either<LaunchRequestArguments, AttachRequestArguments>> ELEMENT_TYPE = new TypeToken<Either<LaunchRequestArguments, AttachRequestArguments>>() {
+	};
+
+	@SuppressWarnings("unchecked")
+	@Override
+	public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
+		Predicate<JsonElement> leftChecker = new PropertyChecker("noDebug");
```
jonahgraham, line 36: Reading this makes it look to me like something is missing from the specification. There is no way to distinguish `launch` and `attach` on the receiving end of this. In https://github.com/microsoft/debug-adapter-protocol/issues/79#issuecomment-1230242253 there is a `request` key that is either `launch` or `attach`, but that isn't explicit in the final spec.

As a result, what is implemented here, basically always creates the `AttachRequestArguments` (because `noDebug` is most likely not present), even if there was a `"request": "launch"` in the JSON. 

I have raised https://github.com/microsoft/debug-adapter-protocol/issues/333

org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/DebugProtocol.xtend
```diff
@@ -4173,3 +4236,19 @@ interface InvalidatedAreas {
...
+class StartDebuggingRequestArguments {
+	/**
+	 * Arguments passed to the new debug session. The arguments must only contain
+   * properties understood by the `launch` or `attach` requests of the debug
+   * adapter and they must not contain any client-specific properties (e.g.
+   * `type`) or client-specific features (e.g. substitutable 'variables').
+	 */
+	@NonNull
+	@JsonAdapter(StartDebuggingRequestArgumentsConfigurationTypeAdapter)
+	Either<LaunchRequestArguments, AttachRequestArguments> configuration;
```
jonahgraham, line 4253: To make this compatible with `launch` and `attach` perhaps it is best to have this be just a `Map<String, Object>`. Otherwise extenders will all have to provide their own extensions of `StartDebuggingRequestArguments` that understands all the fields.

Note that is change would negate the change needed in the type adapter, but I still would like to see that `request` be marked as a required key in that map, at least through documentation.

org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/services/IDebugProtocolClient.java
```diff
@@ -165,7 +167,7 @@ default void process(ProcessEventArguments args) {
 	 * be possible to change that at random times (or too late).
 	 * <p>
 	 * Consequently this event has a hint characteristic: a client can only be
-	 * expected to make a 'best effort' in honouring individual capabilities but
+	 * expected to make a 'best effort' in honoring individual capabilities but
```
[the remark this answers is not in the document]
KamasamaK: My PR microsoft/debug-adapter-protocol#332 was already merged, but not published yet.

org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/adapters/StartDebuggingRequestArgumentsConfigurationTypeAdapter.java
KamasamaK: I noticed that as well, so I didn't bother adding a type adapter for `RestartArguments`, which has the same type. But I guess that was a problem since #618 added it, so I just copied that implementation.

org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/DebugProtocol.xtend
```diff
@@ -4173,3 +4236,19 @@ interface InvalidatedAreas {
...
 	public static final String VARIABLES = "variables";
 }
+
+/**
+ * Arguments for 'startDebugging' request.
+ */
+@JsonRpcData
+class StartDebuggingRequestArguments {
+	/**
+	 * Arguments passed to the new debug session. The arguments must only contain
```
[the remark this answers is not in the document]
KamasamaK: This is fixed. I'm not sure why my editor does that sometimes.

org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/DebugProtocol.xtend
KamasamaK: You're right. I see `launch` and `attach` are not actually using their respective arguments objects. Do you know why that is? Those are the only two requests I see that are like that.

org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/services/IDebugProtocolClient.java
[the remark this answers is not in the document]
jonahgraham: Cool - thanks for doing that upstream too!

org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/adapters/StartDebuggingRequestArgumentsConfigurationTypeAdapter.java
jonahgraham: I see. I guess we'll have to revisit that sometime.

org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/DebugProtocol.xtend
jonahgraham: Because the argument objects aren't ever useful on their own for `launch` and `attach`, you always need extra fields to make them useful. The spec says "Since launching is debugger/runtime specific, the arguments for this request are not part of this specification." Attach says something similar. I suspect that `restart` and `startDebugging` should say something similar. If you agree I will raise that issue (or add it to https://github.com/microsoft/debug-adapter-protocol/issues/333)

In javascript/typescript it can be typed more easily to the `LaunchRequestArguments` because of the language differences (as I am sure you know). I would be open to any idea that made it better here.

jonahgraham (approved):

jonahgraham:
Thanks @KamasamaK I went ahead and merged this following my review so it can be included in 0.16.0 that I am preparing now.

org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/adapters/StartDebuggingRequestArgumentsConfigurationTypeAdapter.java
KamasamaK: Removed this file

PR #679, opened 2022-10-22
Copy Bnd version of Manifest to META-INF. Fixes #678

### LINKED ISSUE #678: Switch from osgi to bnd makes projects that don't work in IDE from source
The steps in [CONTRIBUTING.md#eclipse](https://github.com/eclipse/lsp4j/blob/main/CONTRIBUTING.md#eclipse) no longer create proper OSGi plug-ins in Eclipse anymore since 6a75250. While the steps do create LSP4J projects that can be built and tested in Eclipse, they no longer have their `META-INF/MANIFEST.MF` updated with all the OSGi headers so that Eclipse PDE recognizes it as a proper bundle.

What this means is I can't import both LSP4J and LSP4E as source and do anything as LSP4E can't see LSP4J.

@cdietrich I missed this when testing the gradle upgrade. Any ideas before I debug this more?

cdietrich:
Copy Bnd version of Manifest to META-INF. Fixes #678

@jonahgraham  can you give this one a try?

### DIFF
--- a/gradle/manifest-gen.gradle
+++ b/gradle/manifest-gen.gradle
@@ -23,6 +23,7 @@ jar.bnd (
 	'Bundle-Vendor': 'Eclipse LSP4J',
 	'Bundle-RequiredExecutionEnvironment': 'JavaSE-1.8',
 	"-exportcontents": "org.eclipse.lsp4j.*",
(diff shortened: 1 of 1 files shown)


jonahgraham (approved):
Thanks @cdietrich for figuring this out. This works now.

PR #683, opened 2022-11-16
Fix type adapter for `workspace/diagnostic` response

pisv:
Fixes #682

PR #684, opened 2022-11-16
Update version qualifier for guava

pisv:
Version qualifier for `com.google.guava 30.1.0` has changed in the latest Orbit repository `S20221115203712` from `v20210127-2300` to `v20221112-0806`, which causes LSP4J CI build to fail.


cdietrich (approved):

jonahgraham:
This means we need a new release of LSP4J. Thankfully we don't have a hard dependency in our output on this version, but the p2 repo has the badly signed version and that is being contributed to simrel.

cdietrich:
no issue from Xtext side if the release if available this week. i guess https://github.com/eclipse/lsp4j/pull/683 should also be part

PR #689, opened 2023-01-02
update to gson 2.10

cdietrich:
update to gson 2.10

### DIFF
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -16,8 +16,8 @@ ext.versions = [
 	'xtend_lib': '2.28.0',
 	'guava': '[30.1,31)',
 	'guava_orbit': '30.1.0.v20221112-0806',
-	'gson': '[2.9.1,2.10)',
-	'gson_orbit': '2.9.1.v20220915-1632',
+	'gson': '[2.10,2.11)',
+	'gson_orbit': '2.10.0.v20221207-1049',
 	'websocket_jakarta': '2.0.0',
 	'websocket': '1.0',
 	'junit': '4.13.2'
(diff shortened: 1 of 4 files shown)


gradle/versions.gradle
```diff
@@ -16,8 +16,8 @@ ext.versions = [
 	'xtend_lib': '2.28.0',
 	'guava': '[30.1,31)',
 	'guava_orbit': '30.1.0.v20221112-0806',
-	'gson': '[2.9.1,2.10)',
-	'gson_orbit': '2.9.1.v20220915-1632',
+	'gson': '[2.10,2.11)',
```
HannesWell, line 19:
```suggestion
	'gson': '[2.9.1,2.11)',
```
Gson 2.10 just added new methods (see https://github.com/eclipse-m2e/m2e-core/pull/1146#issuecomment-1362282140). So if lsp4j is using gson as consumer and not as provider, the lower can actually stay where it is.

cdietrich: dont know how versioning works in gson and if using upper range will pick new version in gradle and thus a new method may accidentially be used

cdietrich: at least gradle build scan shows new version picked
https://scans.gradle.com/s/yudurt3xbdgie/dependencies?dependencies=gson&expandAll
am not sure if the 2.10.1 will also drop to orbit.
dont know if and when i will find time to bump

cdietrich: https://git.eclipse.org/r/c/orbit/orbit-recipes/+/199278
but that one will come with m2 only

cdietrich: => propose to merge this one and then do another pr for m2

jonahgraham (approved):

PR #698, opened 2023-02-17
Don't annotate SymbolInformation as Deprecated

### LINKED ISSUE #697: Deprecated SymbolInformation and impact on documentSymbol
Hello lsp4j team:

Related: https://github.com/eclipse/lsp4j/issues/252

A language server of mine does not use `SymbolInformation` directly at all, but it does use  `CompletableFuture<List<Either<SymbolInformation, DocumentSymbol>>> documentSymbol(DocumentSymbolParams)`. 

It would be great if there's a way to avoid the related compile warnings caused by [deprecated `SymbolInformation`](https://github.com/eclipse/lsp4j/blob/3397180c0e163b479ac0524113c76c2070d0b8b8/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend#L6680-L6687).

(In my projects I usually instruct my compiler to treat compile warnings as errors--but that's just a personal preference.)

In my particular (>23K LOC) project the only warnings are happening in my overridden `documentSymbol` method.

I'm aware of the backwards-compatibility considerations, etc., and don't have really definitive suggestions to address this, but can think of perhaps lsp4j also offering an alternative "documentSymbol2" (or some other name/suffix) method only involving `DocumentSymbol`:

```java
@JsonRequest
@ResponseJsonAdapter(DocumentSymbolResponseAdapter.class)
default CompletableFuture<List<DocumentSymbol>> documentSymbol2(DocumentSymbolParams params) {
   throw new UnsupportedOperationException();
}
```

But there are probably other possible ways.

Thanks for considering.

jonahgraham:
While SymbolInformation class is deprecated in the LSP specification. It is not annotated with Deprecated annotation because the TextDocumentService.documentSymbol method requires it and causes deprecated warning in the code of all users of that API.

Instead we mark all the methods/constructors as
@Deprecated so that consumers that are actually using instances of SymbolInformation get deprecated warnings, while avoiding deprecated warnings for just referring to the type.

Fixes #697

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -6679,9 +6679,12 @@ class DocumentSymbol {
 /**
  * Represents information about programming constructs like variables, classes, interfaces etc.
  *
- * @deprecated Use {@link DocumentSymbol} or {@link WorkspaceSymbol} instead if supported.
+ * Deprecated Use {@link DocumentSymbol} or {@link WorkspaceSymbol} instead if supported.
+ *
+ * This class is deprecated in the LSP specification. It is not annotated with Deprecated
+ * annotation because the {@link org.eclipse.lsp4j.services.TextDocumentService.documentSymbol(DocumentSymbolParams)}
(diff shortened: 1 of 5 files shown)


nixel2007:
Sounds reasonable for me.

KamasamaK (approved):

PR #703, opened 2023-03-05
Pass more informative `Throwable` from handler invocations

cmoon4799:
[RemoteEndpoint.java](https://github.com/eclipse/lsp4j/blob/5605ca82252833777dbef82ca13a403dbfb9d906/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/RemoteEndpoint.java#L50) - as it stands, a `Throwable`, deriving from an `InvocationTargetException` from `GenericEndpoint.recursiveFindRpcMethods`, passed into `RemoteEndpoint.exceptionHandler` will not be handled as an `InvocationTargetException`. With this change, a custom `exceptionHandler` will at least be able to use the `Throwable` message that was thrown from the invoked handler.

For example, suppose I have a debug server with a RPC method `Foo` within which an `Exception` is thrown with message `custom string`.  Once `Foo` is invoked and throws an `Exception`, `exceptionHandler` will obtain a `RuntimeException` (from `recursiveFindRpcMethods`) with little connection to the thrown `Exception` - AFAIU, the custom message can only be obtained by parsing the stack trace which may be cumbersome. This change will allow a custom `exceptionHandler` to easily read from `Throwable.message` instead.

### DIFF
--- a/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/GenericEndpoint.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/GenericEndpoint.java
@@ -64,7 +64,7 @@ public class GenericEndpoint implements Endpoint {
 					Object[] arguments = this.getArguments(method, arg);
 					return (CompletableFuture<Object>) method.invoke(current, arguments);
 				} catch (InvocationTargetException | IllegalAccessException e) {
-					throw new RuntimeException(e);
+					throw new RuntimeException(e.getCause());
 				}
 			};
 			if (methodHandlers.put(methodInfo.name, handler) != null) {
--- a/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/GenericEndpoint.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/GenericEndpoint.java
@@ -63,8 +63,10 @@ public class GenericEndpoint implements Endpoint {
 					Method method = methodInfo.method;
 					Object[] arguments = this.getArguments(method, arg);
 					return (CompletableFuture<Object>) method.invoke(current, arguments);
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/GenericEndpoint.java
```diff
@@ -63,7 +63,9 @@ protected void recursiveFindRpcMethods(Object current, Set<Class<?>> visited, Se
 					Method method = methodInfo.method;
 					Object[] arguments = this.getArguments(method, arg);
 					return (CompletableFuture<Object>) method.invoke(current, arguments);
-				} catch (InvocationTargetException | IllegalAccessException e) {
+				} catch (InvocationTargetException e) {
+					throw new RuntimeException(e.getCause());
```
szarnekow, line 67: This will now loose information, e.g. assuming the InvocationTargetException's cause is an IllegalAccessException, the thrown RuntimeException would no longer allow to distinguish the two scenarios.

The use-case you describe in the ticket can be resolved by traversing the causes of the caught RuntimeException in your custom exception handler. You might want to look into  a solution based on `com.google.common.base.Throwables.getCausalChain(Throwable)` or `com.google.common.base.Throwables.getRootCause(Throwable)`

jonahgraham:
Thank you @cmoon4799 for the PR. I think @szarnekow's review probably covers things as for how the change is now. Do you want to provide a PR that better traverses the exceptions in the [DEFAULT_EXCEPTION_HANDLER](https://github.com/eclipse/lsp4j/blob/5605ca82252833777dbef82ca13a403dbfb9d906/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/RemoteEndpoint.java#L50)?

PS To accept PRs we need a signed [ECA](https://github.com/eclipse/lsp4j/blob/main/CONTRIBUTING.md#eca).

jonahgraham:
I am closing this now due to lack of activity and lack of ECA. @cmoon4799 feel free to reopen if you want to address any of the review comments.

Issue #553, opened 2021-05-07
jsonrpc should de-structure byPosition array parameter of requests and notifications


testforstephen:
When i installed vscode-languageclient 7.0.0 to adopt LSP 3.16, i found there was a breaking change in languageclient@7.0.0 about adopting [Parameter Structures](https://www.jsonrpc.org/specification#parameter_structures) concept of jsonrpc@2.0, see [language client 7.0.0 changelog](https://github.com/Microsoft/vscode-languageserver-node#3160-protocol-600-json-rpc-700-client-and-700-server).

> * added support to control the [parameter structure](https://www.jsonrpc.org/specification#parameter_structures) when sending requests and notifications in `vscode-jsonrpc`. The parameter structure can be controlled using the additional `parameterStructures` argument when creating a request or notification type or when sending an untyped request or notification using the `sendRequest` or `sendNotification` function. The default is `ParameterStructures.auto` which does the following:
>    * use `byPosition` for messages with zero or greater than one parameter
>    * for one parameter it used `byName` for parameters which are object literals. Uses `byPosition` for all other parameters.

In summary, if the request/notification parameter is not an object parameter, jsonrpc will wrap it into an array. See [JSON_RPC](https://github.com/microsoft/vscode-languageserver-node/blob/faa68d90a976f158a8403a0f5cce54117ce61dbb/jsonrpc/src/common/connection.ts#L1065-L1083) implementation of language client.

Here are some samples of how to handle parameters in a language client jsonrpc.

- client.sendRequest(type, true)

   jsonrpc:
  {
   ...
   Params: [ true ]
  } 

- client.sendRequest(type, [true])

  jsonrpc:
  {
   ...
   Params: [ [ true ] ]
  }

- client.sendRequest(type, { value: true })

  jsonrpc:
  {
     ...
     Params: {
        value: true
    }
}

To avoid breaking in language server, jsonrpc in lsp4j should de-structure the outermost array wrapper if it‘s a single array parameter.

The code change is parseParams of MessageTypeAdapter.java. If the message parameter is an array, then flatten the array first and map each child item to the parameter type.

https://github.com/eclipse/lsp4j/blob/00447d0a44d75b03b6cebf3d2720a311411efdca/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/MessageTypeAdapter.java#L241-L251

pisv:
[...]

PR #714, opened 2023-03-15
Unwrap single array parameter wrapped by JsonRPC 2.0

dhuebner:
This PR addresses #553 by unwrapping single array parameter.

### DIFF
--- a/org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/json/MessageJsonHandlerTest.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/json/MessageJsonHandlerTest.java
@@ -806,15 +806,20 @@ public class MessageJsonHandlerTest {
 		MessageJsonHandler handler = createSimpleRequestHandler(String.class, new TypeToken<List<Boolean>>() {
 		}.getType());
 
-		var request = new RequestMessage();
-		request.setId(1);
-		request.setMethod(handler.getMethodProvider().resolveMethod(null));
-		request.setParams(List.of(new Boolean[] { true })); // fake wrapped array [[true]] for JsonRpc 2.0
-
-		RequestMessage message = (RequestMessage) handler.parseMessage(request.toString());
+		var request = "{\n"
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/json/MessageJsonHandlerTest.java
```diff
@@ -780,4 +783,61 @@ public void testNotification_AllOrders() {
...
+		request.setId(1);
+		request.setMethod(handler.getMethodProvider().resolveMethod(null));
+		request.setParams(List.of(new Boolean[] { true })); // fake wrapped array [[true]] for JsonRpc 2.0
+
+		RequestMessage message = (RequestMessage) handler.parseMessage(request.toString());
+
+		// Check parse - unwrap array
+		assertEquals(request.getParams(), message.getParams());
+
+	}
```
jonahgraham, line 819: I don't think this test tests what you intended. It is testing `[true]` not `[[true]]` because the `List.of()` overload that is chosen is `List.of(E... elements)`

FWIW I stared at this for a while before I understood what is going on here and realized the unexpected overload was chosen. Perhaps it would be better to use the same style as the other tests and make the JSON explicit? 

```suggestion
	@Test
	public void testArrayParamUnwrap_JsonRpc2_0() {
		MessageJsonHandler handler = createSimpleRequestHandler(String.class, new TypeToken<List<Boolean>>() {
		}.getType());

		var request = "{\n"
				+ "  \"jsonrpc\": \"2.0\",\n"
				+ "  \"id\": 1,\n"
				+ "  \"method\": \"testMethod\",\n"
				+ "  \"params\": [\n"
				+ "    [\n"
				+ "      true\n"
				+ "    ]\n"
				+ "  ]\n"
				+ "}";
		RequestMessage message = (RequestMessage) handler.parseMessage(request);

		// Check parse - unwrap array
		assertEquals(List.of(true), message.getParams());

	}
```

jonahgraham (changes requested):

jonahgraham (approved):

PR #718, opened 2023-03-24
Wrap primitive and array parameter into an array as in JSON-RPC 2.0

dhuebner:
The counter part off my previous PR handling the unwrapping of primitiv and array parameter types as described in  JSON-RPC 2.0.

Added more tests with different parameter types that hopefully makes clear what in wrapped/unwrapped. 

See also:  #553

### DIFF
--- a/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/MessageTypeAdapter.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/MessageTypeAdapter.java
@@ -15,6 +15,7 @@ import java.io.EOFException;
 import java.io.IOException;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
+import java.util.Collection;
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/MessageTypeAdapter.java
```diff
@@ -42,6 +43,7 @@
 import com.google.gson.JsonSyntaxException;
 import com.google.gson.TypeAdapter;
 import com.google.gson.TypeAdapterFactory;
+import com.google.gson.internal.Primitives;
```
jonahgraham, line 46: @dhuebner Can you please have a look at https://github.com/eclipse-lsp4j/lsp4j/issues/738 in relation to this new package dependency?

dhuebner: @jonahgraham 
Sure, will do. Thanks!

Issue #734, opened 2023-05-23
Cannot build with gradle on windows


ghentschke:
I tried to build with gradle on my Windows machine according to this [description](https://github.com/eclipse-lsp4j/lsp4j/blob/main/CONTRIBUTING.md#eclipse) and get the following error during: *4. Run the "eclipse assemble" gradle targets.*:
```java
java.nio.file.InvalidPathException: Illegal char <:> at index 50: C:\git_SC_202212\lsp4j\org.eclipse.lsp4j.jsonrpc\C:git_SC_202212lsp4jorg.eclipse.lsp4j.jsonrpcbuild\tmp\bnd\MANIFEST.MF
	at java.base/sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
	at java.base/sun.nio.fs.WindowsPath.parse(WindowsPath.java:92)
	at java.base/sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:232)
	at java.base/java.io.File.toPath(File.java:2387)
	
```
It seems that the Windows path `C:\git_SC_202212\lsp4j\org.eclipse.lsp4j.jsonrpc` cannot be correctly resolved

cdietrich:
maybe something like
"-savemanifest": "${buildDir.path.replace("\\", "/")}/tmp/bnd/MANIFEST.MF",
helps to get bnd to deal with mix of forward and backwards slashes
or we use something to make the slashes backslashes on windows

jonahgraham:
@ghentschke I have pushed a fix which appears to solve the issue for me, both on CI (which now does Windows builds too #736) and on my local windows machine.

Please let me know if you still run into issues with working on LSP4J from source.

jonahgraham:
> maybe something like `"-savemanifest": "${buildDir.path.replace("\\", "/")}/tmp/bnd/MANIFEST.MF",` helps to get bnd to deal with mix of forward and backwards slashes or we use something to make the slashes backslashes on windows

I couldn't get this to work, not sure what I may have been doing wrong, see #737 for my alternative.

cdietrich:
guess the alternative is also ok as log as the build dir is not changed.
wonder why it worked on my windows.
maybe there is differences between different windows machines and locactions

PR #736, opened 2023-05-29
Run workflow on Windows, macOS and Linux

jonahgraham:
This is to provide a test bench for #734 and to help prevent any fix for #734 from regressing.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -5,7 +5,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - name: Set up JDK 17 ☕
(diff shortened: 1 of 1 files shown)


cdietrich:
@jonahgraham can you check if my patch works with that?

nixel2007:
You could add macos too, cause it may has some os-specific behavior in uri and path handling

jonahgraham:
> @jonahgraham can you check if my patch works with that?

I am looking at that now - I couldn't get your patch to work, but #737 has an alternative I am trying to verify.



> You could add macos too, cause it may has some os-specific behavior in uri and path handling (and yes, lsp4j is used on macos setups as part of bsl language server)

OK, will do.

Issue #738, opened 2023-06-01
Using internals which may not be visible


jonahgraham:
In MessageTypeAdapter we are using an internal Gson type that is exported in [Orbit's gson rebundling (as x-internal)](https://github.com/eclipse/orbit/blob/40c388f997704bbc362af0dfa8c651112bf3bd04/google/com.google.gson_2.10.1/osgi.bnd) but not in the manifest of the upstream gson.

https://github.com/eclipse-lsp4j/lsp4j/blob/e32c63b7b1cea5b777bd0d99c4045f4932ad3ff7/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/MessageTypeAdapter.java#L46

This forces consumers of jsonrpc (within an OSGi runtime) to use the Orbit version.

jonahgraham:
@cdietrich / @dhuebner  looking for a second opinion - is this serious enough that we should do a quick next release? If we (collectively) can fix it immediately then I can do the release ASAP.

cdietrich:
i also would prefer a .1 release otherwise we would need a new xtext release

cdietrich:
i am also not aware of people using lsp4j in osgi context

jonahgraham:
> what i dont understand. why is this not seen in tests. we run non osgi tests 

Without OSGi the LSP4J can access that class (I think? as I don't know how JPMS fits in to the picture, if at all)

---

On more reflection I don't think this is critical, just annoying as it means LSP4J is forcing LSP4E to use the Orbit gson. We should aim to fix soon, but I don't think we need to respin for this.

merks:
It’s not critical.  Next normal release will be soon enough.

jonahgraham:
Thank you @merks for weighing in and giving me that confirmation!

dhuebner:
@jonahgraham 
I can prepare a "fix" for this by copying the  gson implementation of `Primitives.isPrimitive` and `Primitives.isWrapperType`

dhuebner:
@jonahgraham 
I've opened a PR #739

jonahgraham:
> @jonahgraham I can prepare a "fix" for this by copying the gson implementation of `Primitives.isPrimitive` and `Primitives.isWrapperType`

I asked Wayne Beaton review this because I was concerned about the copying but his assessment is in this case it is fine.

PR #739, opened 2023-06-02
Using internals which may not be visible #738

dhuebner:
Removes the `gson.internal.Primitives` import. Added some additional tests. 

See #738.


jonahgraham (approved):

jonahgraham:
I have backported the fix to the 0.21 branch in case I do a 0.21.1 release. See #758

PR #744, opened 2023-07-06
Provide default implementation for stepIn and stepOut redirecting to

### LINKED ISSUE #743: Implement default behavior for stepIn and stepOut to nextStep
Based on [specification for Step In](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_StepIn), `If the request cannot step into a target, stepIn behaves like the next request.`.

Currently, by default the `IDebugProtocolServer.stepIn` [method](https://github.com/eclipse-lsp4j/lsp4j/blob/e32c63b7b1cea5b777bd0d99c4045f4932ad3ff7/org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/services/IDebugProtocolServer.java#L390) is throwing an `UnsupportedOperationException`

I think it could be convenient that it redirect to nextStep by default so that the specification is fulfilled immediately without the client required to implement it

apupier:
next

This is the behavior mentioned in the protocol specification.

fixes #743

### DIFF
--- a/org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/services/IDebugProtocolServer.java
+++ b/org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/services/IDebugProtocolServer.java
@@ -388,7 +388,11 @@ public interface IDebugProtocolServer {
 	 */
 	@JsonRequest
 	default CompletableFuture<Void> stepIn(StepInArguments args) {
-		throw new UnsupportedOperationException();
+		NextArguments nextArgs = new NextArguments();
+		nextArgs.setThreadId(args.getThreadId());
+		nextArgs.setGranularity(args.getGranularity());
+		nextArgs.setSingleThread(args.getSingleThread());
+		return next(nextArgs);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


apupier:
it failed only on Windows with somethign which seems unrelated to my changes:

```
Execution failed for task ':org.eclipse.lsp4j.jsonrpc:test'.
> There were failing tests. See the report at: file:///D:/a/lsp4j/lsp4j/org.eclipse.lsp4j.jsonrpc/build/reports/tests/test/index.html
```

jonahgraham:
> it failed only on Windows with somethign which seems unrelated to my changes:

Passes now on re-run. I'll review soon.

apupier:
up! :-)

(Not urgent but I was checking my list of opened PR)

jonahgraham:
Sorry - fell off my list I guess. Thank you for pinging!

I don't think this change is correct, here are my thoughts

- IDebugProtocolServer provides simply an interface, the behaviour is defined by the adapter itself. All the other methods don't provide such a default. That is consistent with implementation decision in https://github.com/eclipse-lsp4j/lsp4j/tree/main/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services too
- I assume the reason for this change is this line of the spec "If the request cannot step into a target, stepIn behaves like the next request." but that decision about doing a next vs stepIn is normally a context dependent decision by the adapter rather that an always thing. e.g. I don't think anything more than a trivial debugger would work without an implemented stepIn
- the stepOut is not documented to fall back to next, although I suppose that would be an acceptable behaviour if the debugger can tell ahead of time that stepOut won't work. In practice debuggers like gdb can't tell ahead of time the stepOut won't work and will just continue never hitting a stop point

All that said, I don't think that there is much harm in this change, but I don't really want to set a precedent of adding logic into lsp4j.

WDYT?

Issue #763, opened 2023-08-29
There's no need to package `*.java` files into the runtime jar


raylras:
- Including `*.java` files in the runtime jar makes it larger, but they are unnecessary during runtime.
- It causes confusion when browsing the LSP4J source code in IDEA, as it results in duplicate classes.
![image](https://github.com/eclipse-lsp4j/lsp4j/assets/61815515/631877ee-5543-4056-a41b-e94cf825070a)

Solution:
- Simply remove the contents of `org.eclipse.lsp4j/build.gradle` lines 27-29:
```diff
@@ -24,10 +24,6 @@
 	testImplementation "com.tngtech.archunit:archunit:$versions.archunit"
 }
 
-jar {
-	from sourceSets.main.allJava
-}
-
 jar.bnd (
 	'Import-Package': "com.google.gson.*;version=\"$versions.gson\",*"
 )
```

jonahgraham:
Great idea. Can you provide a PR please? We need to also check that the various  sources jars still have all the sources after making the change.

jonahgraham:
Fixed by #764

PR #764, opened 2023-08-29
Remove java files from the runtime jar

raylras:
Related to #763


jonahgraham:
Thank you for the fix @raylras - the next release (0.22.0) expected in the next few weeks will have this fix.

PR #773, opened 2023-11-26
fix: SemanticTokens.getData() returns null for instances via default ctor

sebthom:
`SemanticTokens#getData()` to returns `null` for instances created via default constructor which violates the `@NonNull` value API contract.

This PR is a root cause fix for https://github.com/eclipse/lsp4e/issues/861

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -6184,6 +6184,7 @@ class SemanticTokens {
 	List<Integer> data
 
 	new() {
+		this.data = java.util.Collections.emptyList();
 	}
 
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -6184,6 +6184,7 @@ class SemanticTokens {
 	List<Integer> data
 
 	new() {
+		this.data = java.util.Collections.emptyList();
```
nixel2007, line 6187: You could use import here

pisv:
```suggestion
		this.data = new ArrayList()
```
as in other similar cases in `Protocol.xtend`.

`Collections.emptyList()` cannot be used in this case, as it returns an immutable list.

PR #782, opened 2024-01-10
Update to guava 33.0.0.jre

merks:


merks:
@cdietrich @jonahgraham 

Could someone approve the build?

This change is analogous to the following:

https://github.com/eclipse-lsp4j/lsp4j/commit/523168fb44987087e638ecbfa6640c9f963a774e

releng/releng-target/lsp4j.target.target
```diff
@@ -6,8 +6,8 @@
 			<repository location="https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/2023-09"/>
 			<unit id="com.google.gson" version="2.10.1"/>
 			<unit id="com.google.gson.source" version="2.10.1"/>
-			<unit id="com.google.guava" version="32.1.2.jre"/>
```
cdietrich, line 9: i assume you need to split this up into two orbit locations

cdietrich: https://github.com/eclipse-lsp4j/lsp4j/compare/cd_widenGuava

cdietrich:
plus one. then we can release Xtext with the 0.21.2 dependency in case there wont be a release in lsp4j for 2024-03 and lsp4j can then consume the Xtext 2.34 release

releng/releng-target/lsp4j.target.target
jonahgraham: Why do you think that the orbit should be split up here? I want to make sure I don't mess things up in 0.21.2 that I am working on.

merks: For sure the new version of guava needs 2024-03 to be added...

jonahgraham: I think I answered the question. It is about the renaming of the jakarta.websocket bundles?

merks: Yes, I think these 

image

jonahgraham: Thanks Ed. I think something is going wrong here as we don't use the older jakarta.websocket, but instead jakarta.websocket-api in the code, so I will look at that.

PR #785, opened 2024-01-10
Use/allow latest versions of dependencies for guava and jakarta.websocket

jonahgraham:
A new version of Guava has been released (33) that LSP4J works with and websocket is now published in a "better" way (direct from maven) so we republish that version now.

Consumers shouldn't need to change their dependencies as the lower bounds are unchanged.

cc @merks @cdietrich

### DIFF
--- a/releng/p2/category.xml
+++ b/releng/p2/category.xml
@@ -5,10 +5,12 @@
 	</feature>
 	<bundle id="com.google.gson" version="2.10.1.qualifier"/>
 	<bundle id="com.google.gson.source" version="2.10.1.qualifier"/>
-	<bundle id="com.google.guava" version="32.1.2.qualifier"/>
-	<bundle id="com.google.guava.source" version="32.1.2.qualifier"/>
-	<bundle id="jakarta.websocket" version="0.0.0"/>
-	<bundle id="jakarta.websocket.source" version="0.0.0"/>
+	<bundle id="com.google.guava" version="32.0.0.qualifier"/>
+	<bundle id="com.google.guava.source" version="33.0.0.qualifier"/>
+	<bundle id="jakarta.websocket-api" version="0.0.0"/>
+	<bundle id="jakarta.websocket-api.source" version="0.0.0"/>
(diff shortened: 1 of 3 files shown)


releng/p2/category.xml
```diff
@@ -5,10 +5,12 @@
 	</feature>
 	<bundle id="com.google.gson" version="2.10.1.qualifier"/>
 	<bundle id="com.google.gson.source" version="2.10.1.qualifier"/>
-	<bundle id="com.google.guava" version="32.1.2.qualifier"/>
-	<bundle id="com.google.guava.source" version="32.1.2.qualifier"/>
-	<bundle id="jakarta.websocket" version="0.0.0"/>
-	<bundle id="jakarta.websocket.source" version="0.0.0"/>
+	<bundle id="com.google.guava" version="32.0.0.qualifier"/>
```
pisv, line 8: I know I'm really late with reviewing this, but unfortunately there is a typo in the version number here, which caused the `com.google.guava` bundle to be missing in the p2 update site of the `0.21.2` release...

merks: You'd think the build would fail.  😞 

Fortunately this doesn't affect SimRel because some other repositories provide the needed version...

jonahgraham: @pisv I can do a 0.21.3 very quickly (today) if you think anyone (outside of simrel) is affected. I am hoping that in the next month or so we will end the 0.21.x series and release 0.22.x series from main branch.

> You'd think the build would fail. 😞

:-( I thought I saw it there in the output, but it was just the source bundle and I didn't realize the main bundle was missing or that this wouldn't cause the build to fail.

pisv: @jonahgraham Personally, I think we can do the 0.21.3 release if/when someone else reports the issue.

jonahgraham: I put this fix in place on the branch so we don't forget in https://github.com/eclipse-lsp4j/lsp4j/pull/790

PR #787, opened 2024-01-10
Update CHANGELOG for 0.21.2

jonahgraham:
Part of #784


jonahgraham:
The links in this update will be correct shortly as I am completing the 0.21.2 releng work.

PR #789, opened 2024-01-11
Prepare repo for 0.21.3 development

jonahgraham:


jonahgraham:
I'm not planning on releasing 0.21.3 unless there is any effect, but I wanted to get the fix for missing Guava in p2 site into git as raised by https://github.com/eclipse-lsp4j/lsp4j/pull/785#discussion_r1448607995 - see the fix in #790

PR #795, opened 2024-02-12
fix javadoc warning

jonahgraham:
This pulls out the commit by @cdietrich from https://github.com/eclipse-lsp4j/lsp4j/pull/781

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
@@ -6683,7 +6683,7 @@ class DocumentSymbol {
  * Deprecated Use {@link DocumentSymbol} or {@link WorkspaceSymbol} instead if supported.
  *
  * This class is deprecated in the LSP specification. It is not annotated with Deprecated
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
```diff
@@ -6683,7 +6683,7 @@ class DocumentSymbol {
  * Deprecated Use {@link DocumentSymbol} or {@link WorkspaceSymbol} instead if supported.
  *
  * This class is deprecated in the LSP specification. It is not annotated with Deprecated
- * annotation because the {@link org.eclipse.lsp4j.services.TextDocumentService.documentSymbol(DocumentSymbolParams)}
+ * annotation because the {@link org.eclipse.lsp4j.services.TextDocumentService#documentSymbol(DocumentSymbolParams)}
```
cdietrich, line 6686: am not sure if this also is a warning in 8

cdietrich (approved):

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend
jonahgraham: ok. I think it is more correct with `#` in either case. Thanks for reviewing and approving.

PR #798, opened 2024-02-12
Supply a ToStringBuilder in common location

### LINKED ISSUE #742: Invalid generation of @JsonRpcData when using lsp4j 0.21 and xtend 2.31.0
When installing lsp4j and xtend from the current Eclipse release update site (2023-06) It installs lspj4 0.21.0 and xtend 2.31.0

(I started from Eclipse Modeling 2023-06 and then installed lsp4j using the updatesite https://download.eclipse.org/releases/2023-06/ )

A simple project using @JsonRpcData from lsp4j.generator will now fail.  The generated java class is incorrect. (the toString 

```java
package bug_lsp4j_xtend

import org.eclipse.lsp4j.generator.JsonRpcData

@JsonRpcData
class MyAnnotatedClass {
	
}
```

will generate a code that doesn't compile
image

An equivalent code was working in the previous release I tried (xtend 2.25.0 + lsp4j 0.14.0)  (I haven't tried yet to find the exact last working version ...)

project showing the issue joined
[bug_lsp4j_xtend_2023-06.zip](https://github.com/eclipse-lsp4j/lsp4j/files/11785481/bug_lsp4j_xtend_2023-06.zip)

jonahgraham:
When the dependency on xbase.lib was removed in #529 it put the implicit requirement that there is a class called ToStringBuilder in a sub-package of the xtend class called `util`.

This change provides the ToStringBuilder in an util package in the org.eclipse.lsp4j.jsonrpc bundle so that consumers of LSP4J do not need to copy ToStringBuilder into their own source code.

Fixes #742


jonahgraham:
@eclipse-lsp4j/technology-lsp4j-committers if someone has time to review this and weigh in on whether it should make it into 0.22.0 that I will build tomorrow that will be great.

org.eclipse.lsp4j/.settings/org.eclipse.core.resources.prefs
```diff
@@ -1,7 +1,6 @@
 eclipse.preferences.version=1
 encoding//src/main/java=UTF-8
 encoding//src/main/resources=UTF-8
-encoding//src/main/xtend-gen=UTF-8
```
pisv, line 4: Not quite sure about this change, it looks a bit incidental...

Anyway, it should not hurt, since `encoding/<project>=UTF-8` covers it all.

Just wondering what was the reason for this change.

pisv (commented):
@jonahgraham Thank you very much for taking care of it!

The patch looks fine to me and safe enough to release it sooner rather than later 👍

org.eclipse.lsp4j/.settings/org.eclipse.core.resources.prefs
jonahgraham: Thanks for noticing - I accidentally squashed that change into this PR. I have separated it out to a new PR now.

PR #799, opened 2024-02-12
Document the LSP4J JSON-RPC generator a little

jonahgraham:
Documented as per #798


jonahgraham:
If #798 doesn't get merged soon I will update this to document without #798 included.

PR #801, opened 2024-02-13
Supply a Preconditions in common location

jonahgraham:
This change provides the Preconditions in an util package in the org.eclipse.lsp4j.jsonrpc bundle so that consumers of LSP4J do not need to copy Preconditions into their own source code.

Related to #742


jonahgraham:
(Oops - I missed something in this push, fix incoming)

jonahgraham:
I am closing this - I prefer getting rid of, rather than deprecating the "old" Preconditions anyway.

Issue #647, opened 2022-07-09
javax.websocket


cdietrich:
build uses javax.websocket from orbit, which is no longer available in newer orbit versions.
we might have to use a newer orbit version for a newer gson eventually.

=> how to resolve? is jakarta.websocket a feasible alternative?

pisv:
> => how to resolve? is jakarta.websocket a feasible alternative? so that we can drop the javax one?

I don't think so, see #536. Could not we just stick to the last available javax.websocket from Orbit?

cdietrich:
I think we can, just makes the target file more complicated / less straight forward

jonahgraham:
Note that this is a subissue of #874 to remove all deprecated API for 1.0 release. PR is ready to review (along with rest of deprecated API removal) in #881

PR #803, opened 2024-02-13
Deprecate lsp4j.websocket in preparation for removal in the future.

jonahgraham:
The bundle has been annotated deprecated for removal in a future version.

The bundle will not be part of the SDK feature, nor part of SimRel going forward.

Part of #647


jonahgraham:
I want to verify the CI built output on Jenkins which will be in https://download.eclipse.org/lsp4j/builds/jonah/deprecate-javax once Jenkins is working again. See https://github.com/eclipse-lsp4j/lsp4j/issues/732#issuecomment-1941774706

PR #809, opened 2024-02-14
Unwrap InvocationTargetException which are ResponseErrorException

### LINKED ISSUE #802: ResponseErrorException is not propagated with the default exception handler
The same as: https://github.com/eclipse-lsp4j/lsp4j/issues/576
But by then only the doc had changed: https://github.com/eclipse-lsp4j/lsp4j/pull/578

In my opinion this is not very user friendly. If I bother to throw a `ResponseErrorException`, it should be propagated even if it is thrown when the method is called and not only when the `CompletableFuture` is executed.

As an example, we do check some context stuff that is only available as a `ThreadLocal` and not while the `CompletableFuture` is executed.

There is also an open source example on [eclpse-gslp](https://github.com/eclipse-glsp/) that does this incorrectly. See [DefaultGLSPServer](https://github.com/eclipse-glsp/glsp-server/blob/2bb9e0b545f9f2657b26c8526f16a40ef5d2228e/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/protocol/DefaultGLSPServer.java#L77)

jonahgraham:
By wrapping ResponseErrorException with another layer in RuntimeException it means the unwrapping of the exception in RemoteEndpoint.DEFAULT_EXCEPTION_HANDLER doesn't work as expected and instead of getting the original ResponseError the other end gets the fallback ResponseError of type internal error.

This change includes updates to documentation to show that it is ok to throw ResponseErrorException (reverts https://github.com/eclipse-lsp4j/lsp4j/pull/578). While it is also OK to return an exceptionally completed future, it is not needed to do that way and simply throwing is more straightforward.

There are new tests to cover the changed handling. Also tested is the previously documented way of handling exceptions (see `tries == 1` case in `testVersatility` and `testVersatilityResponseError`

Fixes https://github.com/eclipse-lsp4j/lsp4j/issues/802

### DIFF
--- a/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/GenericEndpoint.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/GenericEndpoint.java
@@ -21,7 +21,6 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
 import java.util.function.Function;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -60,14 +59,20 @@ public class GenericEndpoint implements Endpoint {
 		AnnotationUtil.findRpcMethods(current.getClass(), visited, (methodInfo) -> {
 			@SuppressWarnings("unchecked")
 			Function<Object, CompletableFuture<Object>> handler = (arg) -> {
+				Method method = methodInfo.method;
+				Object[] arguments = this.getArguments(method, arg);
 				try {
-					Method method = methodInfo.method;
-					Object[] arguments = this.getArguments(method, arg);
 					return (CompletableFuture<Object>) method.invoke(current, arguments);
 				} catch (InvocationTargetException e) {
-					throw new CompletionException(e.getCause());
+					Throwable cause = e.getCause();
+					if (cause instanceof RuntimeException) {
+						throw (RuntimeException) cause;
+					} else if (cause instanceof Error) {
+						throw (Error) cause;
+					}
+					throw new IllegalStateException("An unexpected exception occurred while executing jsonrpc method " + method, cause);
 				} catch (IllegalAccessException e) {
-					throw new CompletionException(e);
+					throw new IllegalStateException("Inaccessible jsonrpc method: " + method, e);
 				}
 			};
 			if (methodHandlers.put(methodInfo.name, handler) != null) {
@@ -179,5 +184,4 @@ public class GenericEndpoint implements Endpoint {
 	protected boolean isOptionalMethod(String method) {
 		return method != null && method.startsWith("$/");
 	}
-
 }
--- a/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/GenericEndpoint.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/GenericEndpoint.java
@@ -1,5 +1,5 @@
 /******************************************************************************
- * Copyright (c) 2016 TypeFox and others.
+ * Copyright (c) 2016, 2024 TypeFox and others.
  * 
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -21,6 +21,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.function.Function;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -63,8 +64,10 @@ public class GenericEndpoint implements Endpoint {
 					Method method = methodInfo.method;
 					Object[] arguments = this.getArguments(method, arg);
 					return (CompletableFuture<Object>) method.invoke(current, arguments);
-				} catch (InvocationTargetException | IllegalAccessException e) {
-					throw new RuntimeException(e);
+				} catch (InvocationTargetException e) {
+					throw new CompletionException(e.getCause());
+				} catch (IllegalAccessException e) {
+					throw new CompletionException(e);
 				}
 			};
 			if (methodHandlers.put(methodInfo.name, handler) != null) {
--- a/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/GenericEndpoint.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/GenericEndpoint.java
@@ -83,7 +83,7 @@ public class GenericEndpoint implements Endpoint {
 					LOG.fine("A delegate object is null, jsonrpc methods of '" + method + "' are ignored");
 				}
 			} catch (InvocationTargetException | IllegalAccessException e) {
-				throw new RuntimeException(e);
+				throw new IllegalStateException("An exception occurred while executing JsonDelegate method " + method, e);
 			}
 		});
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 Fixed issues: <https://github.com/eclipse-lsp4j/lsp4j/milestone/34?closed=1>
 
   * The exception handling around throwing `ResponseErrorException` has been improved to ensure that it is unwrapped to the expected `ResponseError` on the receiving side.
+  In addition, `@JsonDelegate`s that throw exceptions have their checked exceptions wrapped in the more narrow `IllegalStateException` instead of a `RuntimeException`.
     * See [#802](https://github.com/eclipse-lsp4j/lsp4j/issues/802) for detailed discussion.
 
 Breaking API changes:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 Fixed issues: <https://github.com/eclipse-lsp4j/lsp4j/milestone/34?closed=1>
 
+  * The exception handling around throwing `ResponseErrorException` has been improved to ensure that it is unwrapped to the expected `ResponseError` on the receiving side.
+    * See [#802](https://github.com/eclipse-lsp4j/lsp4j/issues/802) for detailed discussion.
+
 Breaking API changes:
 
 Nightly japicmp report: <https://download.eclipse.org/lsp4j/builds/main/japicmp-report/>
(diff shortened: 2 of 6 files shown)


pisv (commented):
@jonahgraham This looks like a nice improvement indeed, thank you! 👍

It seems that the documentation update in #578 needs to be reverted or amended now, either as part of this PR or a follow-up.

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/GenericEndpoint.java
```diff
@@ -63,7 +63,13 @@ protected void recursiveFindRpcMethods(Object current, Set<Class<?>> visited, Se
 					Method method = methodInfo.method;
 					Object[] arguments = this.getArguments(method, arg);
 					return (CompletableFuture<Object>) method.invoke(current, arguments);
-				} catch (InvocationTargetException | IllegalAccessException e) {
+				} catch (InvocationTargetException e) {
+					if (e.getTargetException() instanceof ResponseErrorException) {
```
ivy-cst, line 67: Looks good, just some notes:

This could be a breaking change for existing exception handlers.

If the behavior is changed we could even throw all target exceptions which are an instance of `RuntimeException` directly, without wrapping?
E.g. 

```
if (e.getTargetException() instanceof RuntimeException) {
  throw (RuntimeException) e.getTargetException();
}
```
And may be even remove always the `InvocationTargetException` which is a technical detail because reflection is used?
```
if (e.getTargetException() instanceof RuntimeException) {
  throw (RuntimeException) e.getTargetException();
}
throw new RuntimeException(e.getTargetException());
```

pisv: @ivy-cst Just wondering, why do you think this could be a breaking change for existing exception handlers? I assume that any exception handlers should have been prepared to handle a `ResponseErrorException` thrown directly, in the same way as `DEFAULT_EXCEPTION_HANDLER` does.

In general, I think that it would have been better to use a `CompletionException` as a wrapper rather than a `RuntimeException` here, or perhaps even ["sneaky throws"](https://www.baeldung.com/java-sneaky-throws) without any wrapping at all.

jonahgraham: I'm not worried about this being a breaking change - I don't think that we document what is expected beyond what `DEFAULT_EXCEPTION_HANDLER` does already and the current `DEFAULT_EXCEPTION_HANDLER` can handle these new cases.

I do agree that wrapping in a `CompletionException` and discarding the `InvocationTargetException` is a reasonable thing to do.

I see that `DEFAULT_EXCEPTION_HANDLER` has a bug itself regarding `InvocationTargetException`. It has this code:

```java
if ((throwable instanceof CompletionException || throwable instanceof InvocationTargetException)
				&& throwable.getCause() instanceof ResponseErrorException)
```

but in the case of InvocationTargetException this will never be true because InvocationTargetException's cause is always null, and thus never an instanceof `ResponseErrorException`. 

So in conclusion more updates are needed. I'll revisit soon, but if someone want to take this up in the meantime that would be great because my allocated LSP4J time is running out quick.

jonahgraham: > I have implemented an alternative approach in a separate commit, so it can easily be reviewed and reverted if necessary.

I'll have a look.

> The Javadoc for InvocationTargetException.getTargetException() says that getCause() is now the preferred means of obtaining the target exception, so this code looks OK.

Sorry - that was misinformation on my part. Thanks for checking and sorry for the noise.

jonahgraham: Lets go with wrapping in CompletionException as that looks good. I will squash your change when I next push this PR.

jonahgraham:
> There is one todo thing left, hence turned to draft, the code change when it is delegate segments has no test coverage. Probably also worth adding to changelog.

Newest commit adds to the changelog. It also adds the code coverage as the find delegate's change I had done previously was incorrect.

ivy-cst (approved):
Looks good, thanks for the fix :+1:

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/services/GenericEndpoint.java
```diff
@@ -63,8 +64,10 @@ protected void recursiveFindRpcMethods(Object current, Set<Class<?>> visited, Se
 					Method method = methodInfo.method;
 					Object[] arguments = this.getArguments(method, arg);
 					return (CompletableFuture<Object>) method.invoke(current, arguments);
-				} catch (InvocationTargetException | IllegalAccessException e) {
-					throw new RuntimeException(e);
+				} catch (InvocationTargetException e) {
+					throw new CompletionException(e.getCause());
+				} catch (IllegalAccessException e) {
+					throw new CompletionException(e);
 				}
```
pisv, line 71: I have some second thoughts about this change.

The reason is that this code is used for handling both requests and notifications.

While using `CompletionException` as a wrapper makes sense for requests, it seems much less natural for notifications.

In fact, any exceptions thrown from notification handlers are logged at

https://github.com/eclipse-lsp4j/lsp4j/blob/88868d3506d317b12cf0bfd1e82a254489f8c47b/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/RemoteEndpoint.java#L233

and discarded, since notifications are not expected to send anything back. In our case, the logged exception would be the wrapping `CompletionException`, which might seem a bit confusing. We could, of course, expect the `CompletionException` and unwrap it before logging, but it looks more complicated than necessary and odd.

So, I wonder: would using sneaky throw instead of throwing `CompletionException` a more consistent approach for both requests and notifications?

jonahgraham: I don't have a strong opinion on this - the important thing is that a `ResponseError` gets unwrapped. I don't know how important the intermediary exceptions are.

As of now I have never used sneaky throws, so don't know if they would be suitable. If you think they are, I'm very ok to defer to your experience.

From https://github.com/eclipse-lsp4j/lsp4j/pull/809#discussion_r1491277086:

> [...] if someone want to take this up in the meantime that would be great because my allocated LSP4J time is running out quick.

I can probably find a bit more time, but I hope that someone else can help this come to a conclusion :-)

pisv: I'll try the sneaky throw approach and see how it goes then. I agree that it probably does not matter much, provided that the main goal of this PR is achieved, but it might be worthwhile to try different approaches, especially since we now have great test coverage around this code area -- thank you :-)

ivy-cst: In my opinion sneaky throws looks like a hack and seems to work by accident in java.

pisv: > In my opinion sneaky throws looks like a hack and seems to work by accident in java.

This looks like a strong opinion. But are you speaking in general or do you see any specific drawbacks in using it in this particular case?

ivy-cst: No one expects a checked exception if it is not specified.
We can simply wrap the exception in a specific `RuntimeException` and then unwrap it before calling the exception handler. This way we can hide the technical detail that we need to wrap it internally.

pisv: > No one expects a checked exception if it is not specified.

As far as I can see, the exception handler expects a `Throwable` as its argument, which includes any checked 
exception.

Also, looking at the implementation of `RemoteEndpoint.handleRequest` and `handleNotification`, I cannot see any drawbacks with using sneaky throws in this case.

> We can simply wrap the exception in a specific RuntimeException and then unwrap it before calling the exception handler. This way we can hide the technical detail that we need to wrap it internally.

Then, we would need to make the `RemoteEndpoint` aware of this specific runtime exception type used in the `GenericEndpoint`, which does not look like a good design to me.

pisv: Overall, I agree with you that sneaky throws is not a general purpose tool and should not be used in ordinary code. But for the right job, in essentially reflective code, it can be a valuable tool. It is just based on how the JVM works, not on how the Java compiler works.

ivy-cst: I agree. May be checked exceptions are anyway very rare. So if we only wrap in that case it would may be also ok?

```
if (e.getCause() instanceof RuntimeException) {
  throw (RuntimeException) e.getCause();
}
throw new CompletionException(e.getCause());
```

pisv: Sure, this would also be a valid approach. It is just that I'm looking for something that would work a bit more uniformly/consistently across requests/notifications, checked/unchecked exceptions. I have just tried the sneaky throws approach locally and it seems to deliver what I'm looking for. In my experience, it is the right approach in this case. Anyway, thank you for discussing this and other alternatives.

pisv:
The 'sneaky throws' change is ready for review :-)

pisv:
I have another approach in the works, which does not use sneaky throws. It is based on the fact that an `Endpoint` is not expected to throw a checked exception. Therefore, throwing a checked exception from an annotated jsonrpc method can (and should) be treated as a programming error, and can be signalled via an `IllegalStateException` in the `GenericEndpoint`. An inaccessible jsonrpc method (`IllegalAccessException`) can also be handled in a similar way. There is no need to use sneaky throws or a `CompletionException` in this case. Runtime exceptions or errors would just be re-thrown.

pisv:
The latest change is ready for review: https://github.com/eclipse-lsp4j/lsp4j/pull/809/commits/3982fd3ba22c2ae71cebafdc3647dbfd347ff770.

jonahgraham (commented):
(Silly GitHub - I can't approve my own PR even though I am reviewing @pisv's latest change)

+1 on this change.

pisv (approved):
Given that @jonahgraham has approved my latest change, this PR is complete and ready to be merged, as far as I can tell. 

Thank you @jonahgraham and @ivy-cst!

jonahgraham:
@ivy-cst this change should end up on snapshot builds soon: https://github.com/eclipse-lsp4j/lsp4j?tab=readme-ov-file#snapshots

Please comment in #807 if you have input on if/when a full release will help your adoption so that I can consider how and when to make a release.

jonahgraham:
OK thanks for responding - if that changes please comment on #807 

PS I really appreciate that you followed up on this issue, especially knowing that it did not affect your code in the short term. It will now make everyone's code in the future better and is part of what makes me very happy to be working on open source projects with a vibrant community!

PR #816, opened 2024-02-29
Avoid temporary String object creation in StreamMessageProducer

sebthom:
This can reduce the likeliness of OOMs when dealing with large response messages from LS servers. See discussion at https://github.com/eclipse-lsp4j/lsp4j/issues/815

PR #817, opened 2024-03-01
Refactoring/code cleanup

sebthom:
To work on #816 I imported the lsp4j project into my Eclipse workspace and was greeted with a lot of compiler and style warnings. With this PR I am addressing a few of the issues. Each type of change is encapsulated in a separate commit for easier review.

If you disagree with some changes I can remove the respective commits.

### DIFF
--- a/org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/adapters/RestartArgumentsArgumentsTypeAdapter.java
+++ b/org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/adapters/RestartArgumentsArgumentsTypeAdapter.java
@@ -1,17 +1,15 @@
 /******************************************************************************
  * Copyright (c) 2022 itemis AG and others.
- * 
+ *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0,
  * or the Eclipse Distribution License v. 1.0 which is available at
  * http://www.eclipse.org/org/documents/edl-v10.php.
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
  ******************************************************************************/
 package org.eclipse.lsp4j.debug.adapters;
-
-import java.util.function.Predicate;
 
 import org.eclipse.lsp4j.debug.AttachRequestArguments;
 import org.eclipse.lsp4j.debug.LaunchRequestArguments;
@@ -20,7 +18,6 @@ import org.eclipse.lsp4j.jsonrpc.json.adapters.EitherTypeAdapter.PropertyChecker
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
 
 import com.google.gson.Gson;
-import com.google.gson.JsonElement;
 import com.google.gson.TypeAdapter;
 import com.google.gson.TypeAdapterFactory;
 import com.google.gson.reflect.TypeToken;
@@ -33,10 +30,9 @@ public class RestartArgumentsArgumentsTypeAdapter implements TypeAdapterFactory 
 	@SuppressWarnings("unchecked")
 	@Override
 	public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
-		Predicate<JsonElement> leftChecker = new PropertyChecker("noDebug");
-		Predicate<JsonElement> rightChecker = leftChecker.negate();
-		TypeAdapter<Either<LaunchRequestArguments, AttachRequestArguments>> elementTypeAdapter = new EitherTypeAdapter<>(
-				gson, ELEMENT_TYPE, leftChecker, rightChecker);
+		final var leftChecker = new PropertyChecker("noDebug");
+		final var rightChecker = leftChecker.negate();
+		final var elementTypeAdapter = new EitherTypeAdapter<>(gson, ELEMENT_TYPE, leftChecker, rightChecker);
 		return (TypeAdapter<T>) elementTypeAdapter;
 	}
 
--- a/org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/adapters/RestartArgumentsArgumentsTypeAdapter.java
+++ b/org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/adapters/RestartArgumentsArgumentsTypeAdapter.java
@@ -27,7 +27,7 @@ import com.google.gson.reflect.TypeToken;
 
 public class RestartArgumentsArgumentsTypeAdapter implements TypeAdapterFactory {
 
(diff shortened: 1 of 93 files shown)


org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/LauncherTest.java
```diff
@@ -46,14 +46,12 @@ static class Param {
 		public String message;
 	}
 	
-	static interface A {
-		@JsonNotification
-		public void say(Param p);
+	interface A {
+		@JsonNotification void say(Param p);
```
pisv, line 50: Not sure about the reformatting that happened at this and other similar places in this commit. Was it intentional?

org.eclipse.lsp4j.jsonrpc.debug/src/main/java/org/eclipse/lsp4j/jsonrpc/debug/json/DebugMessageJsonHandler.java
```diff
@@ -30,7 +30,8 @@ public DebugMessageJsonHandler(Map<String, JsonRpcMethod> supportedMethods, Cons
 		super(supportedMethods, configureGson);
 	}
 
-	public GsonBuilder getDefaultGsonBuilder() {
+	@Override
+   public GsonBuilder getDefaultGsonBuilder() {
```
pisv, line 34: Indentation seems to be amiss at this and other places across this commit.

org.eclipse.lsp4j.debug/src/main/java/org/eclipse/lsp4j/debug/adapters/RestartArgumentsArgumentsTypeAdapter.java
```diff
@@ -33,10 +30,9 @@ public class RestartArgumentsArgumentsTypeAdapter implements TypeAdapterFactory
 	@SuppressWarnings("unchecked")
 	@Override
 	public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
-		Predicate<JsonElement> leftChecker = new PropertyChecker("noDebug");
-		Predicate<JsonElement> rightChecker = leftChecker.negate();
-		TypeAdapter<Either<LaunchRequestArguments, AttachRequestArguments>> elementTypeAdapter = new EitherTypeAdapter<>(
-				gson, ELEMENT_TYPE, leftChecker, rightChecker);
+	   final var leftChecker = new PropertyChecker("noDebug");
+	   final var rightChecker = leftChecker.negate();
+	   final var elementTypeAdapter = new EitherTypeAdapter<>(gson, ELEMENT_TYPE, leftChecker, rightChecker);
```
pisv, line 35: Indentation seems to be amiss at this and other places across this commit.

pisv (commented):
@sebthom Thanks a lot for putting the effort into cleaning up and modernizing the code base, looks really impressive! 👍 

There are a few minor formatting-related remarks, and I'm not a big fan of the last commit that introduces `var`s, but that's just my personal opinion. Otherwise, LGTM.

BTW, it was a pleasure to review this PR. Thanks again!

jonahgraham:
I'm fine with this - but was confused because I didn't see these warnings, until I realized that the projects were defaulting to workspace settings and my workspace and yours must be different.

Can you please consider adding a commit where you save the compiler warnings/etc and code style setting to the project?

Without your change, in my workspace the only warnings I see are deprecated warnings.

pisv:
> Without your change, in my workspace the only warnings I see are deprecated warnings.

Same here.

BTW, there is also `@SuppressWarnings("all")` on the `interface Tuple`, which seems completely unnecessary and was suppressing the warnings about unused imports.

sebthom:
@pisv 
> I'm not a big fan of the last commit that introduces `var`s, but that's just my personal opinion

I am aware that this is a sensitive topic :-) On the one hand I like code that is explicit on the other hand I like code that is terse. To get the best of both worlds I therefore only use the `var` keyword when the actual type is still visible on the right side of the assignment either via an explicit cast `var foo = (SomeType) obj' or if it is a constructor call `var foo = new SomeType()'. In those cases the providing the type on the left side is really redundant. However I am not a fan of relying on type interference in other cases like `var foo = someService.someMethod().get(2)`.

org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/LauncherTest.java
sebthom: This is from the commit that removes redundant modifiers. So it is intended. Methods declared on interfaces are always public.

pisv:
@sebthom  Regarding introducing `var`s into the code base, your approach certainly makes sense to me, thanks for the detailed description.

It is just that I have grown very much accustomed to always seeing the type on the left of the variable declaration in the good old Java days :-) So, it is just my personal preference, and I understand that I'm quite biased towards it. Besides, I'm currently spending much more time with TypeScript, and slowly getting the type inference thing :-)

Therefore, I'd like to leave this decision to others. Since @jonahgraham as the project lead apparently has no objections regarding it, I think we are good to go with it 👍

org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/LauncherTest.java
pisv: (Strange, I have received your comment by email, but GitHub seems tor refuse displaying it here.)

By the reformatting, I meant the annotation being moved from its own line.

jonahgraham:
> Since @jonahgraham as the project lead apparently has no objections

No objections :-)

FWIW your journey sounds just like mine re vars and living in typescript land more and more.

sebthom:
> Can you please consider adding a commit where you save the compiler warnings/etc and code style setting to the project?

@jonahgraham I added my Eclipse compiler settings to the build.gradle. If you regenerate the eclipse settings using `gradle eclipse` all projects will be adjusted accordingly. Fell free to change them.

jonahgraham (approved):
LGTM - lets start with the setting as you have them. If we run into problems we can change later.

At some point we should do the same for formatting and other style issues such as the fixed trailing whitespace. (I'm not going to get around to that, but welcome anyone else who wants to put in the effort)

@pisv Please let me know if you approve before I submit.

@sebthom to make your life easier, would you like to see this PR or #816 merged first?

org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/LauncherTest.java
pisv: Let's keep it that way. It's just a nitpick, anyway.

pisv (approved):
Looks great to me 👍 Wholeheartedly approve, and much appreciate this PR.

pisv:
> FWIW your journey sounds just like mine re vars and living in typescript land more and more.

After having lived in the Eclipse IDE land for almost 20 years, I started contributing to Eclipse Theia about a year ago :-)

PR #818, opened 2024-03-01
Add dependabot config and update GHA workflow

sebthom:


jonahgraham (approved):

jonahgraham:
Thanks @sebthom for this - we had a couple of PRs come in from dependabot now.

PR #821, opened 2024-03-16
Fix small typo

yuk1ty:
`need to to do` -> `need to do` :D


pisv (approved):
@yuk1ty Thank you for taking heed 👍

jonahgraham:
Thanks @yuk1ty for the fix and @pisv for the merge. It is great to know that these resources are useful to the community.

PR #825, opened 2024-04-04
Add `reason` field to Breakpoint

Soarex16:
According to the specification, the Breakpoint type must have `reason` optional field. This field was introduced in microsoft/debug-adapter-protocol#437.


jonahgraham:
Thanks for the submission. `reason` was added in 1.64 and LSP4J has only implemented up until 1.60 so far.

I'm therefore not sure what to do with this pull request. Hopefully someone can implement 1.61-1.64 (or 65) so I don't need to decide how to handle this special case.

Would you like to submit a PR to upgrade to 1.64 or 1.65?
-

jonahgraham:
That is great @Soarex16 - I look forward to reviewing it when you are ready.

PR #837, opened 2024-05-21
broaden gson range, bundle new gson and guava

cdietrich:
@jonahgraham do we need a dot release for simrel?

### DIFF
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -15,7 +15,8 @@ version = '0.24.0-SNAPSHOT'
 ext.versions = [
 	'xtend_lib': '2.32.0',
 	'guava': '[32.1.2,34)',
-	'gson': '[2.9.1,2.12)',
+	'gson': '[2.9.1,3.0)',
+
 	'websocket_jakarta': '2.0.0',
 	'websocket': '1.0',
 	'junit': '4.13.2',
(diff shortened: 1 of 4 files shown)


szarnekow (approved):

szarnekow:
As a client of lsp4j I'd even consider using a 3.0 as the exclusive upper bound of the dependency.

jonahgraham:
> As a client of lsp4j I'd even consider using a 3.0 as the exclusive upper bound of the dependency.

We got (minorly) burned a while ago because a new version of gson didn't strictly follow semantic versioning. Ever since then I have been concerned to not have a runtime problem.

However I am asked multiple times for this, so seeing as many consumers seem ok with the (minor) risk, I say lets widen it up.

> @jonahgraham do we need a dot release for simrel?

Yes

gradle/versions.gradle
```diff
@@ -15,7 +15,7 @@ version = '0.24.0-SNAPSHOT'
 ext.versions = [
 	'xtend_lib': '2.32.0',
 	'guava': '[32.1.2,34)',
-	'gson': '[2.9.1,2.11)',
+	'gson': '[2.9.1,2.12)',
```
jonahgraham, line 18:
```suggestion
	'gson': '[2.9.1,3.0)',

```

cdietrich:
Can you merge am afk

PR #845, opened 2024-08-18
Support inactive regions

ghentschke:
this feature is needed for a clangd protocol extension. See this cdt-lsp issue https://github.com/eclipse-cdt/cdt-lsp/issues/299 and this clangd commit: https://reviews.llvm.org/D143974 It is needed for macro enable in the C/C++ language


ghentschke:
Not sure if this should/must be done in LSP4J

pisv:
@ghentschke Thank you for taking the time to submit this contribution.

> Not sure if this should/must be done in LSP4J

As long as it is not in the LSP spec (and, as far as I can see, it is not in the current 3.17 version or the upcoming 3.18 version), it should not be done in LSP4J. If you think that it should become part of the spec, you can try to contribute this protocol extension to https://github.com/microsoft/language-server-protocol. Then, it could be supported by LSP4J...

PR #860, opened 2024-11-30
Update build settings for changes to JIPP as a result of recent LTS update

cdietrich:
See https://www.eclipse.org/lists/cross-project-issues-dev/msg19990.html for announcement


jonahgraham:
Thanks @cdietrich for starting this work. I'll see if I can get it working now.

jonahgraham:
I've done my best to test this without merging, but the final gpg part needs a merge, so I am going to do that and patch up as needed.

Issue #866, opened 2025-01-26
ClassCastException Either3 cannot be cast to InlineValue when handling "textDocument/inlineValue" event


sebthom:
I am trying to implement partial InlineValue support for LSP4E but I am hitting a ClassCastException.
I believe the problem is that TextDocumentService.inlineValue has a return type of `CompletableFuture<List<InlineValue>>` 
https://github.com/eclipse-lsp4j/lsp4j/blob/651eb49e1f4cd90439808885ddaad0862b330908/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/TextDocumentService.java#L676-L679
but the InlineValueResponseAdapter defines element type as `Either3<InlineValueText, InlineValueVariableLookup, InlineValueEvaluatableExpression>` instead of `InlineValue` :
https://github.com/eclipse-lsp4j/lsp4j/blob/651eb49e1f4cd90439808885ddaad0862b330908/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/adapters/InlineValueResponseAdapter.java#L32-L33

pisv:
Turns out that #867 was not a complete fix for the issue, as the element type adapter in `InlineValueResponseAdapter` still created instances of `Either3` rather than `InlineValue` after the fix. Unfortunately, `testInlineValueResponseN` methods in `JsonParseTest` (erroneously) expected instances of `Either3` instead of `InlineValue` and were not fixed appropriately, so the issue had not been caught up then.

I have created #985 to fix it.

pisv:
Reopening the issue as per the previous comment.

PR #867, opened 2025-01-26
fix: ClassCastException when handling "textDocument/inlineValue" event

sebthom:
Addresses #866


pisv (approved):
LGTM 👍 My apology for this bug, and thanks a lot for fixing it up!

sebthom:
@pisv do you think a new lsp4j release can be created in the foreseeable future?

pisv:
@sebthom As far as I know, https://github.com/eclipse-lsp4j/lsp4j/issues/834#issuecomment-2578257566 is the current information on this topic. If you have input on when the new release would align well with your schedule, you can comment on https://github.com/eclipse-lsp4j/lsp4j/issues/834.

In the meantime, the latest integration build with the fix is available at https://download.eclipse.org/lsp4j/builds/main/.

PR #882, opened 2025-02-12
update guava and gson

cdietrich:


jonahgraham:
I assume this update isn't needed for 2025-03?

cdietrich:
Depends on how many goons we want to end up in simrel

jonahgraham (approved):
We don't have dups in SimRel for these at the moment. This change LGTM, but it only affects what we build and publish, it doesn't change our dependencies so won't affect what is in SimRel.

When SimRel / Maven Central has new versions of guava and gson than in our [versions gradle file](https://github.com/eclipse-lsp4j/lsp4j/blob/main/gradle/versions.gradle), we can adjust the ranges.

So, if I am not missing anything, I won't do a new release for this change.

cdietrich:
lets see what m3 will show.

PR #885, opened 2025-04-01
Omit missing params when writing request/notification messages

### LINKED ISSUE #655: JSON-RPC request without params violates specification with null
When a JSON-RPC request without `params` (i.e. them being `null` in Java) is emitted, that gets written as `"params": null` instead of just omitting the `params` field from the output. The latter should be done instead.

## Specifications
[JSON-RPC specification](https://www.jsonrpc.org/specification#request_object) says:

> **params**
>    A _Structured value_ that holds the parameter values to be used during the invocation of the method. This member MAY be omitted. 

where "two structured types (Objects and Arrays)".

[LSP specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#requestMessage) says (in TypeScript syntax):

> ```
> /**
>  * The method's params.
>  */
> params?: array | object;
> ```

which exactly matches the JSON-RPC specification with the `?` meaning omission, but not `null` (which would have to be explicitly given as a `| null` variant).

### Implementation
The writing of `params` here:
https://github.com/eclipse/lsp4j/blob/22f8edce55379bd9d4ae1b96bc614e74549bc0f9/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/MessageTypeAdapter.java#L404-L409
uses the `writeNullValue` function here:
https://github.com/eclipse/lsp4j/blob/22f8edce55379bd9d4ae1b96bc614e74549bc0f9/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/MessageTypeAdapter.java#L449-L457
which goes out of its way to output a `null` instead of omitting missing `params`.

This clearly violates the specifications and trips up other JSON-RPC/LSP implementations that require incoming requests to follow the specification to the letter.

### Background
According to git blame, this behavior was introduced in #137, the goal of which was to fix the behavior for response `result` that indeed may not be omitted but may be `null`. Looks like the same logic was erroneously applied to request/notification `params` as well.

pisv:
Fixes #655.

### DIFF
--- a/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/MessageTypeAdapter.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/MessageTypeAdapter.java
@@ -418,12 +418,11 @@ public class MessageTypeAdapter extends TypeAdapter<Message> {
 			writeId(out, requestMessage.getRawId());
 			out.name("method");
 			out.value(requestMessage.getMethod());
-			out.name("params");
 			Object params = requestMessage.getParams();
-			if (params == null)
-				writeNullValue(out);
-			else
+			if (params != null) {
+				out.name("params");
 				handleParameter(out, params, requestMessage.getMethod());
+			}
 		} else if (message instanceof ResponseMessage) {
 			final var responseMessage = (ResponseMessage) message;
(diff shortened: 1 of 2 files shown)


jonahgraham (approved):
LGTM - but due to previous comments if @cdietrich has time to look that would be great.

sebthom:
Can this be merged please. 🙏

PR #889, opened 2025-05-04
added missing about.htmls

cdietrich:


merks (approved):
Very good.  Thank you.

cdietrich:
ci is still down so for this to be shown we will need to wait for the gitlab issue to be worked on

PR #899, opened 2025-10-25
fix: correct ResourceOperationTypeAdapter dispatch and add guard

sebthom:
- Replace reversed `isAssignableFrom` checks with `instanceof` to correctly handle subclasses (e.g., `CreateFile` derivatives) in write(...)
- Add explicit `JsonParseException` for unknown `ResourceOperation` subtypes in write(...) to mirror read(...) and avoid silent malformed JSON
- Improves protocol correctness and failure transparency when encountering unsupported operations

Tests: add `ResourceOperationTypeAdapterTest` covering subclass serialization and unknown subtype failure

### DIFF
--- a/org.eclipse.lsp4j/src/test/java/org/eclipse/lsp4j/test/adapters/ResourceOperationTypeAdapterTest.java
+++ b/org.eclipse.lsp4j/src/test/java/org/eclipse/lsp4j/test/adapters/ResourceOperationTypeAdapterTest.java
@@ -16,15 +16,14 @@ import static org.junit.Assert.assertTrue;
 
 import org.eclipse.lsp4j.CreateFile;
 import org.eclipse.lsp4j.ResourceOperation;
+import org.eclipse.lsp4j.adapters.ResourceOperationTypeAdapter;
 import org.junit.Test;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 
 /**
- * Failing tests that demonstrate issues in ResourceOperationTypeAdapter:
- * - reversed isAssignableFrom dispatch prevents handling subclasses
- * - missing guard for unknown subtypes silently produces invalid output
+ * Tests for {@link ResourceOperationTypeAdapter}.
  */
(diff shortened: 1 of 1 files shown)


pisv:
@sebthom Good catch, thank you for the fix! 👍

It is already a great improvement, but would not it even better to check that the `type` passed to `ResourceOperationTypeAdapter.create` is assignable to `CreateFile` or `DeleteFile` or `RenameFile`? This would make it unnecessary to throw `JsonParseException` from `InnerResourceOperationTypeAdapter.write`.

(I don't quite like throwing `JsonParseException` from `InnerResourceOperationTypeAdapter.write`. According to its Javadoc, this exception is raised if there is a serious issue that occurs during *parsing* of a Json string. If anything, I'd prefer to throw `AssertionError` in this case.)

sebthom:
This does not work `type.getRawType()` is always ResourceOperation.class and never one of the sub classes. I verified that via the debugger.

pisv:
It does have a practical relevance by setting a precedent of throwing `JsonParseException` in write methods in the project's code base. By doing so, it sets an example that probably will be copied in the future.

Otherwise, we could just as well throw a `RuntimeException` (or any subclass of it) in this case.

sebthom:
Maybe this should be discussed separately and addressed in a separate PR? I can see other places where `JsonParseException` is thrown in a similar manner, like in EitherTypeAdapter.create or MessageTypeAdapter.createMessage. So if this is changed than probably better everywhere at once.

pisv:
Thank you for discussing it in detail. The only reason for this discussion is that I could find no prior cases of throwing `JsonParseException` from write (serialization) methods neither in GSON itself nor in LSP4J. (The methods you mentioned where `JsonParseException` is thrown are only called from read methods.) That's why I'm concerned that this PR is setting a precedent in throwing `JsonParseException` from a write method.

pisv:
I'd suggest throwing `IllegalArgumentException` instead of `JsonParseException` in this case. So, yes, please. Thank you!

sebthom:
@pisv done

PR #900, opened 2025-10-25
build: update GHA workflow config

sebthom:
- update URLs to reference documentation
- test on MacOS ARM and Intel
- update versions of github actions (checkout, setup-java)
- parameterize validate job so that in case of errors an SSH debug session can be opened
- decouple Gradle JRE from compile/test JDK


jonahgraham (changes requested):
I have split this out into #918 which includes these changes plus some others, in particular to allow build and test with Java 21 while still generating Java 11 bytecode.

PR #902, opened 2025-10-25
perf: eliminate intermediate byte[] copies in StreamMessageConsumer

sebthom:
- Add MessageJsonHandler.serialize(Message, OutputStream, Charset)
- Serialize into ByteArrayOutputStream and write via writeTo(output)
- Remove String.getBytes(...) and toByteArray() clone
- Cache Charset instead of using encoding String

No breaking changes: existing constructors retained; new overloads are additive.

### DIFF
--- a/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/MessageJsonHandler.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/MessageJsonHandler.java
@@ -12,10 +12,13 @@
 package org.eclipse.lsp4j.jsonrpc.json;
 
 import java.io.IOException;
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
 import java.io.Reader;
 import java.io.StringReader;
 import java.io.StringWriter;
 import java.io.Writer;
+import java.nio.charset.Charset;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


pisv:
> I don't see how to provide realistic benchmarks.

OK. But have you measured the actual increase in performance somehow?

pisv:
Thanks for looking into this, Jonah! I really appreciate it 👍

As hinted in https://github.com/eclipse-lsp4j/lsp4j/pull/902#issuecomment-3522424622, it did not seem so clear-cut to me. The new code allocates a byte-buffer of size `8192` for a `StreamEncoder` of the `OutputStreamWriter` per each message, and the `StreamEncoder` creates a new char array and wraps it into a new `CharBuffer` in each `write` call, which can hinder the expected gain in eliminating intermediate `byte[]` copies.

However, my general point has been that any performance-related PRs should be justified by their contributors with at least some description of the actual issue they observed in practice as evidenced by profiling results and at least some measurements of the actual performance increase.

For example, when contributing to Eclipse Theia, they *require* of each PR to include the "How to test" section that should explain how a reviewer can reproduce a bug, test new functionality or verify performance improvements (*sic*). This is enforced by [the PR template](https://github.com/eclipse-theia/theia/blob/master/.github/PULL_REQUEST_TEMPLATE.md?plain=1).

To put it bluntly, committers should not be expected to do all the work we are doing now to try and justify each attempt at micro-optimization, which usually just is not worth the effort, in my experience.

Issue #423, opened 2020-03-30
Cancellation does not work after CancellableFuture transformation


tpasternak:
The cancellation mechanism of `RemoteEndpoint` requests is implemented in that way, the returned `CompletableFuture` contains an overriden `cancel` method, that sends cancel request to the server.

Unfortunately, `CompletableFuture` loses this information after transformations like `thenApply` or `thenCompose`, so the user has to remember that cancellation must be done on the original Future.

This code demonstrates the `CompletableFuture` problem https://gist.github.com/tpasternak/b2a460fb0bcde6a26a7f81f2b57a3b63

PR #907, opened 2025-10-26
feat(jsonrpc): add JsonRpcRequestFuture with explicit protocol cancel

sebthom:
Return a `JsonRpcRequestFuture` from `RemoteEndpoint.request(...)`, enabling `cancelRequest(boolean)` to send `$/cancelRequest` exactly once per request chain.

Semantics are kept predictable:
- Root future `cancel(boolean)` delegates to `cancelRequest(...)` (sends protocol cancel).
- Derived stages do not auto-send on `cancel(boolean)`; callers use `cancelRequest(...)` explicitly.

----
This PR addresses #423 and would also greatly simplify future handling in LSP4E.

You can now do `endpoint.request("foo", "myparam").thenApply(..).thenCompose(...).thenAccept(...).cancelRequest(true)` to cancel the complete future chain. I believe this also makes the cumbersome `org.eclipse.lsp4j.jsonrpc.CancelChecker`/ `org.eclipse.lsp4j.jsonrpc.FutureCancelChecker` in a lot of cases obsolete.

### DIFF
--- a/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/JsonRpcRequestFuture.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/JsonRpcRequestFuture.java
@@ -94,9 +94,9 @@ public final class JsonRpcRequestFuture<T> extends CompletableFuture<T> {
 	@Override
 	public boolean cancel(final boolean mayInterruptIfRunning) {
 		// Root cancel ensures cancelRequest() behavior once before local cancellation
-		if (isRoot()) {
+		if (root == this) {
 			try {
-				sendCancelOnceIfNeeded();
+				cancelRequest();
 			} catch (final RuntimeException ex) {
 				// catching potential exception to ensure local cancel still proceeds
 				LOG.log(Level.WARNING, ex.getMessage(), ex);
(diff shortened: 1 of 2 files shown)


pisv:
Thank you for this PR! 👍

I quite like the design in general, but I'd assume that calling `cancelRequest` on a dependent future should also cancel the root future. However, it does not seem to be the case. If I add `assertTrue(original.isCancelled());` to the end of `testCancellationAfterTransformation`, the test fails. Am I missing something?

sebthom:
I changed it accordingly. 

`future.cancelRequest()` now will only send the `$/cancelRequest` but not immediately cancel any futures locally. Instead the future will be cancelled indirectly via processing of the cancel response from the server. 

Additionally `root.cancel(...)` will call `this.cancelRequest() + super.cancel(...)`.

pisv:
Great! Thank you! 👍

sebthom:
@pisv I addressed this via https://github.com/eclipse-lsp4j/lsp4j/pull/909

pisv:
@sebthom Thanks for looking into the issue.

As far as I understand, attempting to cancel a request is inherently racy with regard to the request completion. For example, the request might complete before the cancel notification arrives. This is normal, and should be expected.

Actually, attempting to cancel an ordinary `CompletableFuture` is also inherently racy with regard to the future completion.

Therefore, I can see no problem with the code of this PR in this regard. In other words, I don't think that the changes proposed in #909 are really necessary or even desired.

Perhaps the test itself expects too much, let's try to take a deeper look at it.

Issue #593, opened 2022-01-20
WebSocket message consumer incompatible with TracingMessageConsumer


zxdxjtu:
In WebSocketLauncherBuilder, WebSocketMessageConsumer is assigned to TracingMessageConsumer by MessageTracer.apply

```
MessageConsumer outgoingMessageStream = new WebSocketMessageConsumer(session, jsonHandler);
outgoingMessageStream` = wrapMessageConsumer(outgoingMessageStream);
```
But in MessageTracer.consume, it only consumes StreamMessageConsumer and RemoteEndpoint class, WebSocketMessageConsumer is incompatible

zxdxjtu:
Throws `Unknown MessageConsumer type` when using WebSocketLauncherBuilder
```
               if (messageConsumer instanceof StreamMessageConsumer) {
			logString = consumeMessageSending(message, now, date);
		} else if (messageConsumer instanceof RemoteEndpoint) {
			logString = consumeMessageReceiving(message, now, date);
		} else {
			LOG.log(WARNING, String.format("Unknown MessageConsumer type: %s", messageConsumer));
			logString = null;
		}
```

PR #908, opened 2025-10-26
fix(tracing): stop 'Unmatched response message' logs for non-stream IO

sebthom:
This PR addresses #593

Previously only StreamMessageConsumer was treated as "sending". With WebSocket and other non-stream transports, sentRequests stayed empty, so every incoming response looked unmatched, emitted a warning, and produced no trace.

Now any consumer that is not a RemoteEndpoint is treated as "sending", ensuring sentRequests is populated regardless of transport and responses are matched correctly. Removed the "Unknown MessageConsumer type" branch to avoid misclassification and improve compatibility with custom/non-stream transports.

### DIFF
--- a/org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/TracingMessageConsumerTest.java
+++ b/org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/TracingMessageConsumerTest.java
@@ -11,12 +11,8 @@
  ******************************************************************************/
 package org.eclipse.lsp4j.jsonrpc.test;
 
-import org.eclipse.lsp4j.jsonrpc.*;
-import org.eclipse.lsp4j.jsonrpc.TracingMessageConsumer.RequestMetadata;
-import org.eclipse.lsp4j.jsonrpc.json.MessageJsonHandler;
-import org.eclipse.lsp4j.jsonrpc.json.StreamMessageConsumer;
-import org.eclipse.lsp4j.jsonrpc.messages.*;
-import org.junit.Test;
+import static java.util.Collections.emptyMap;
+import static org.junit.Assert.*;
 
 import java.io.ByteArrayOutputStream;
 import java.io.File;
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/TracingMessageConsumerTest.java
```diff
@@ -299,6 +299,72 @@ public void testSendingNotificationWithCustomAdapter() {
 
 		assertEquals(expectedTrace, actualTrace);
 	}
+
+	@Test
+	public void testUnmatchedResponseWhenOutgoingIsNotStreamConsumer() {
```
pisv:
```suggestion
	public void testNoUnmatchedResponseWhenOutgoingIsNotStreamConsumer() {
```

org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/TracingMessageConsumerTest.java
```diff
@@ -299,6 +299,72 @@ public void testSendingNotificationWithCustomAdapter() {
...
+		var outgoingTrace = new StringWriter();
+		var outgoingPrintWriter = new PrintWriter(outgoingTrace);
+
+		var incomingTrace = new StringWriter();
+		var incomingPrintWriter = new PrintWriter(incomingTrace);
+
+		var sentRequests = new HashMap<String, RequestMetadata>();
+		var receivedRequests = new HashMap<String, RequestMetadata>();
+
+		// Wrap a non-stream consumer so TracingMessageConsumer does not treat it as "sending"
```
pisv: This comment looks a bit misleading, as it describes the old implementation rather than the new one.

org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/TracingMessageConsumerTest.java
```diff
@@ -299,6 +299,72 @@ public void testSendingNotificationWithCustomAdapter() {
...
+		var outgoingTracing = new TracingMessageConsumer(new EmptyMessageConsumer(), sentRequests, receivedRequests, outgoingPrintWriter, TEST_CLOCK_1, Locale.US);
+
+		// Send a request (this will NOT populate sentRequests because consumer is not StreamMessageConsumer)
```
pisv: This comment also describes the old implementation rather than the new one. I'd just remove it, as it can be misleading.

pisv (approved):
LGTM! Thank you for the fix 👍

PR #909, opened 2025-10-31
fix(jsonrpc): cancel locally first and still emit $/cancelRequest once

sebthom:
- Prevent a race where a server's RequestCancelled response completes the future with ResponseErrorException before local cancellation, breaking expected CancellationException behavior.
- Change cancel(boolean) to cancel the future locally first, then (root-only) send $/cancelRequest.
- Adjust cancelRequest() to emit once even if the future is already completed locally by removing the isDone() short-circuit; dedup remains via cancelSent.


pisv:
@sebthom As mentioned in https://github.com/eclipse-lsp4j/lsp4j/pull/907#issuecomment-3476755474, I don't think that we should try to fix what is not broken. Therefore, I propose to close this PR in favour of #910, if you don't mind. Thanks again for looking into this issue, it is much appreciated!

PR #915, opened 2025-11-04
Add utility method for request cancellation

pisv:
This PR adds a utility method to cancel the JSON-RPC request associated with a `CompletableFuture`. It shall allow clients to avoid downcasts to `JsonRpcRequestFuture` in their code.

It is a follow-up to #907.

cc @sebthom


pisv:
@sebthom  Thanks for the suggestion. I think it will not hurt, although I could not imagine a practical use-case for checking the return value, to be honest.

I have updated this PR accordingly.

sebthom:
LGTM!

PR #921, opened 2025-11-18
Document LSP4J's API policy

jonahgraham:
Fixes https://github.com/eclipse-lsp4j/lsp4j/issues/873

Note that I want to enforce the policy too via using `--error-on-semantic-incompatibility` for japicmp, but that doesn't seem to be supported from CLI so I put forward https://github.com/siom79/japicmp/pull/467 and if that is accepted we can automate the missing version bump detection

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,3 +96,41 @@ LSP4J is published under two licenses:
  * [Eclipse Distribution License v. 1.0](https://www.eclipse.org/org/documents/edl-v10.php), a [BSD-3-clause license](https://opensource.org/licenses/BSD-3-Clause)
 
 SPDX-License-Identifier: `EPL-2.0 OR BSD-3-Clause`
+
+### API Policy
+
+The Eclipse LSP4J project uses Semantic Versioning to determine the version number increment the project should use, as determined by [japicmp](https://siom79.github.io/japicmp/CliTool.html).
+All bundles within LSP4J share a version number to simplify version management, both for LSP4J project as well as API consumers of LSP4J.
+Therefore the version increment for any LSP4J bundle will be the largest increment required by semantic versioning of any of the bundles.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -96,3 +96,40 @@ LSP4J is published under two licenses:
...
+
+The Eclipse LSP4J project uses Semantic Versioning to determine the version number increment the project should use, as determined by [japicmp](https://siom79.github.io/japicmp/CliTool.html).
+All bundles within LSP4J share a version number to simplify version management, both for LSP4J project as well as API consumers of LSP4J.
+Therefore the version increment for any LSP4J bundle will be the largest increment required by semantic versioning of any of the bundles.
+
+#### New versions of LSP and DAP specifications
+
+New versions of [LSP](https://microsoft.github.io/language-server-protocol/) and [DAP](https://microsoft.github.io/debug-adapter-protocol/)
+specifications may lead to breaking API changed due to mapping of TypeScript types to Java types.
+Where possible, the LSP4J project will avoid making breaking changes when adapting to new releases of the protocols.
```
pisv: I'd probably suggest to omit this sentence. When adapting to new releases of the protocols, conforming to the protocol in the most direct way possible seems to be more important than trying to avoid breaking changes, even if they could potentially be avoided by some intricate mapping of the protocol.

README.md
```diff
@@ -96,3 +96,40 @@ LSP4J is published under two licenses:
...
+
+It is particularly difficult to maintain binary compatibility when adopting to some protocol changes that allow return values or fields to have new/additional types.
```
pisv:
```suggestion
It is particularly difficult to maintain binary compatibility when adapting to some protocol changes that allow return values or fields to have new/additional types.
```

PR #929, opened 2025-11-23
Add JMH benchmarks to jsonrpc

jonahgraham:
This commit adds some basic benchmarks to serve as a starting point for future benchmarking work.

On the CI we build the benchmarks, but we don't run them.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -65,7 +65,7 @@ jobs:
         cache: gradle
 
     - name: Build with Gradle 🏗️
-      run: ./gradlew build testOlderJavas
+      run: ./gradlew build jmhCompileGeneratedClasses testOlderJavas
 
     - name: Upload Test Results
       if: always()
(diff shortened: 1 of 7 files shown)


jonahgraham:
@sebthom / @pisv I have made some benchmarks so that we have a basis for further discussions. Please have a look at your convenience.

sebthom:
It is very useful to have some benchmark infrastructure like this ready to use. However regarding JMH I'd like to point out that it is GPL licensed. I once tried to add JMH benchmarks to TM4E but it was rejected by the project lead for legal concerns at the time. Maybe better check with Eclipse Legal team first.

jonahgraham:
The IP team have previously reviewed and accepted use of JMH, ref: [iplab search results](https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/?sort=created_date&state=closed&search=maven%2Fmavencentral%2Forg.openjdk.jmh%2Fjmh-core%2F%2a&first_page_size=20) - we _may_ not be able to redistribute jmh parts, but since we don't want to do that, we are ok to proceed here.

PR #932, opened 2025-11-23
Remove all warnings in lsp4j

jonahgraham:
See individual commits + commit messages for details of why and how different warnings are handled.

Fixes https://github.com/eclipse-lsp4j/lsp4j/issues/923

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -132,7 +132,7 @@ subprojects {
 			properties['org.eclipse.jdt.core.compiler.problem.pessimisticNullAnalysisForFreeTypeVariables']='warning'
 			properties['org.eclipse.jdt.core.compiler.problem.possibleAccidentalBooleanAssignment']='warning'
 			properties['org.eclipse.jdt.core.compiler.problem.potentialNullReference']='warning'
-			properties['org.eclipse.jdt.core.compiler.problem.potentiallyUnclosedCloseable']='warning'
+			properties['org.eclipse.jdt.core.compiler.problem.potentiallyUnclosedCloseable']='ignore'
 			properties['org.eclipse.jdt.core.compiler.problem.rawTypeReference']='warning'
 			properties['org.eclipse.jdt.core.compiler.problem.redundantNullAnnotation']='warning'
 			properties['org.eclipse.jdt.core.compiler.problem.redundantNullCheck']='warning'


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/NotebookDocumentService.java
```diff
@@ -24,7 +23,7 @@ public interface NotebookDocumentService {
 	/**
 	 * The open notification is sent from the client to the server when a notebook document
 	 * is opened. It is only sent by a client if the server requested the synchronization
-	 * mode {@code notebook} in its {@link ServerCapabilities#notebookDocumentSync} capability.
+	 * mode {@code notebook} in its {@code ServerCapabilities.notebookDocumentSync} capability.
```
pisv: Not sure whether it'd worth the effort, but potentially we could use constructs like this in such cases:

```suggestion
	 * mode {@code notebook} in its {@link ServerCapabilities#getNotebookDocumentSync() notebookDocumentSync} capability.
```

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/EitherTypeAdapter.java
```diff
@@ -210,13 +210,19 @@ protected Either<L, R> create(JsonToken nextToken, JsonReader in) throws IOExcep
 		throw new JsonParseException("Unexpected token " + nextToken + ": expected " + left + " | " + right + " tokens.");
 	}
 
+	/**
+	 * @throws IOException
```
pisv: Not sure about suppressing these unused throws warnings in this way. I think that at least `Either.createLeft/createRight` should not throw `IOException`, but this probably also applies to `EitherTypeArgument.read(JsonElement)` and `StreamMessageProducer.handleMessage`.

jonahgraham: I thought about that too - but I didn't want to introduce an API breaking change to fix these warnings. If we want to do that, it should be pulled out into a different PR and be documented in API breaking changes.

org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/NotebookDocumentService.java
jonahgraham: That is a good solution. I will do that. I hadn't wanted `{@link ServerCapabilities#getNotebookDocumentSync()}` alone, because it looked "bad". Bug something like you have, although probably as below instead, I think is ok.
```suggestion
	 * mode {@code notebook} in its {@link ServerCapabilities#getNotebookDocumentSync() ServerCapabilities.notebookDocumentSync} capability.
```

org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/adapters/EitherTypeAdapter.java
jonahgraham: OK - will do.

org.eclipse.lsp4j.jsonrpc/src/test/java/org/eclipse/lsp4j/jsonrpc/test/annotations/impl/GenericEndpointTest.java
```diff
@@ -97,6 +97,7 @@ public void testSimple() {
 		Assert.assertEquals(2, foo.calls);
 	}
 
+	@SuppressWarnings("unused")
```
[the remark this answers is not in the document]
pisv: Unfortunately, I still have these problems in a fresh clone...
Screenshot 2025-11-23 at 20 45 54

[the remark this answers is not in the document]
szarnekow: Mind that the most recent JDT milestone has fixed a bug in the interpretation of deprecated members/classes. That might cause the confusion here.

PR #954, opened 2026-02-09
Add read/write access tags to call hierarchy and find references (fix #951)

### LINKED ISSUE #951: Add read and write tags to "find references" and "call hierarchy" responses (adapt API to proposal from LSP PR 2226)
When showing the results of a [_find references_](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_references) or [_call hierarchy_](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_prepareCallHierarchy) request, I'd like the UI to display if a reference or method call in the result is accessing an element (e.g. a variable) by reading or writing it or both. For that purpose, an [LSP specification PR](https://github.com/microsoft/language-server-protocol/pull/2226) suggests to introduce new tags and to adapt the _find references_ response's return type.

For example, the call hierarchy view in CDT (classic C/C++ editor) shows overlay icons with an "R" for read access, "W" for write access, or "RW" for both (see screenshot and see original LSP issue https://github.com/microsoft/language-server-protocol/issues/2207)

Image

travkin79:
Add new tags as proposed in LSP specification issue https://github.com/microsoft/language-server-protocol/issues/2207 and PR https://github.com/microsoft/language-server-protocol/pull/2226 to add read / write access details to [call hierarchy](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_prepareCallHierarchy) and [find reference](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_references) responses. Fix #951.

### DIFF
--- a/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/TextDocumentService.java
+++ b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/TextDocumentService.java
@@ -73,6 +73,7 @@ import org.eclipse.lsp4j.PrepareRenameDefaultBehavior;
 import org.eclipse.lsp4j.PrepareRenameParams;
 import org.eclipse.lsp4j.PrepareRenameResult;
 import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4j.Reference;
 import org.eclipse.lsp4j.ReferenceParams;
 import org.eclipse.lsp4j.RenameParams;
 import org.eclipse.lsp4j.SelectionRange;
@@ -218,6 +219,22 @@ public interface TextDocumentService {
 	 */
 	@JsonRequest
 	default CompletableFuture<List<? extends Location>> references(ReferenceParams params) {
+		throw new UnsupportedOperationException();
+	}
+	
+	/**
+	 * <p>The references request is sent from the client to the server to resolve
+	 * project-wide references for the symbol denoted by the given text document
+	 * position.</p>
+	 * 
+	 * Registration Options: {@link org.eclipse.lsp4j.ReferenceRegistrationOptions}
+	 * 
+	 * <p>This is an LSP <b>proposal</b>. See <a href="https://github.com/microsoft/language-server-protocol/pull/2226">PR</a>
+	 * This method is planned to replace {@link #references(ReferenceParams)} and could be renamed to 'references' in future.</p>
+	 */
+	// TODO introduce this new method (avoid a breaking change) or replace #references(ReferenceParams)?
+	@JsonRequest
+	default CompletableFuture<Either<List<Location>,List<Reference>>> referencesWithTags(ReferenceParams params) {
 		throw new UnsupportedOperationException();
 	}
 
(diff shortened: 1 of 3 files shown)


org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/TextDocumentService.java
```diff
@@ -220,6 +221,22 @@ default CompletableFuture<Either<List<? extends Location>, List<? extends Locati
...
+	 * <p>The references request is sent from the client to the server to resolve
+	 * project-wide references for the symbol denoted by the given text document
+	 * position.</p>
+	 * 
+	 * Registration Options: {@link org.eclipse.lsp4j.ReferenceRegistrationOptions}
+	 * 
+	 * <p>This is an LSP <b>proposal</b>. See <a href="https://github.com/microsoft/language-server-protocol/pull/2226">PR</a>
+	 * This method is planned to replace {@link #references(ReferenceParams)} and could be renamed to 'references' in future.</p>
+	 */
+	// TODO introduce this new method (avoid a breaking change) or replace #references(ReferenceParams)?
```
travkin79: Hi @jonahgraham,
I'm still not sure how we should handle the new return type alternatives from the [proposed LSP change](https://github.com/microsoft/language-server-protocol/pull/2226/changes#diff-c5bc620ba862d66c1d27889dcf8c6be569f9a58442342566107ffa39d58d6ac5R73). We could, of course, change the return type in `references(ReferenceParams)` instead of introducing a new method `referencesWithTags(ReferenceParams)` and increase LSP4J's version according to semantic versioning, but since the proposed LSP change is not yet accepted, I'm not feeling comfortable with introducing a breaking change that we might have to roll back if the LSP proposal gets rejected.

How do you think about introducing this additional method instead? It would avoid introducing a breaking change and even if the LSP proposal would be rejected some day, LSP4J wouldn't have to change the original method signature in `references(ReferenceParams)`.

jonahgraham: If we agree that LSP4J is the reference implementation (in a similar way https://github.com/microsoft/language-server-protocol/pull/2003 is for symbol tags) then the "correct" version should be implemented here.

Which means making the API breaking change which is explicitly allowed in the API policy of the project https://github.com/eclipse-lsp4j/lsp4j?tab=readme-ov-file#new-versions-of-lsp-and-dap-specifications

IMHO it is cleaner to change the API than to start coding in workarounds.

If the change is reverted, we bump the version number again.

jonahgraham (commented):
Until there is more sign that this is a likely to be accepted addition to the LSP I don't think LSP4J should release it. The Symbol Tags proposal (https://github.com/microsoft/language-server-protocol/pull/2003) which has been merged in LSP4J has some indication that it may be approved because of real engagement from the LSP maintainers.

That said, I don't want to end up in a Catch-22 situation on progressing this, so please let us know if LSP4J needs to adopt this to get things unlocked.

PR #970, opened 2026-03-23
Fix ResponseErrorCode casing in shutdown method example

JonathanStarup:
This should be uppercase, matching: https://github.com/eclipse-lsp4j/lsp4j/blob/5486d6fc35c87b22afd0a241154ca27f204dd9da/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/messages/ResponseErrorCode.java#L57-L61


pisv (approved):
Thanks!

Issue #775, opened 2023-12-15
Server side LSP request hangs if executed from a notification handler


MiklosMagyari:
I don't know if it is by design, but at least it was not trivial for me and took several hours to track down.

We have a language server built on lsp4j. Among many others, we implemented the handler for `didChangeWorkspaceFolders`. After handling the folder changes, we call the function performing a project build that creates a work progress:

    @Override
    public void didChangeWorkspaceFolders(DidChangeWorkspaceFoldersParams params) {
        ...
        build();  // at a point it sends work done progress create
        ...
    }

It turned out that calling the build like this causes  sending `window/workDoneProgress/create` request to hang. Checking the trace in vscode showed that the request has been sent out and vscode responded almost immediatelly, but the CompletableFuture returned by `createProgress()` did not complete.

Finally, I have changed the code to:

    @Override
    public void didChangeWorkspaceFolders(DidChangeWorkspaceFoldersParams params) {
        ...
        CompletableFuture.runAsync(() -> {
            build();
        });
        ...
    }

This way I have no lockup.

Is it intentional that I practically need to exit notification handlers before sending out an LSP request?

pisv:
@MiklosMagyari Thank you for the detailed report.

Basically, LSP4J runs a loop in a dedicated thread that reads messages from an input stream and dispatches them to the corresponding handlers (for details, see `StreamMessageProducer`). Therefore, it is never a good idea to implement the handlers for requests or notifications in a blocking, synchronous way.

In your specific case, the deadlock occurred because the thread that reads and dispatches incoming messages was blocked by synchronously waiting for the response to the `window/workDoneProgress/create` request and, hence, was not able to read that response in the first place.

HTH

MiklosMagyari:
Thanks a lot for the clarification, that explains why this lockup happened.

Btw, is this documented somewhere? Maybe I have missed it?
I believe this information would be useful for others as well.

pisv:
@MiklosMagyari I have opened PR #777 to update the documentation.

Any comments are welcome!

MiklosMagyari:
Great! Short and concise, can spare hours of debugging. Thanks once again.

PR #975, opened 2026-06-15
Fix minor typos and clarify non-blocking handler guidance

Kaike-blanck:
Summary

This PR includes three small documentation and cleanup changes:

Fixes a typo in the warning message for unknown message types in RemoteEndpoint.
Fixes a typo in the Javadoc of TracingMessageConsumer.
Clarifies the recommendation for request and notification handlers to avoid blocking the message dispatching thread.

Motivation

These changes are minor, but they improve readability and documentation clarity. The added note in the documentation makes the existing recommendation about non-blocking handlers more explicit by suggesting the use of CompletableFuture for long-running work.

Impact

This PR does not change runtime behavior or public APIs. The changes are limited to a log message typo, Javadoc text, and documentation.

### DIFF
--- a/documentation/README.md
+++ b/documentation/README.md
@@ -43,6 +43,8 @@ This will start the listening process in a new thread that reads messages from t
 
 When implementing the handlers for requests or notifications, you need to be aware that the calling thread is the thread that reads and dispatches incoming messages. Therefore, blocking it may result in reduced throughput or even a deadlock (https://github.com/eclipse-lsp4j/lsp4j/issues/775). As a general rule, message handlers should be implemented in a non-blocking, asynchronous way.
 
+For long-running work, prefer returning a `CompletableFuture` and performing the actual computation outside of the message dispatching thread. This keeps the JSON-RPC message processor responsive while the request is being handled asynchronously.
+
 To stop listening for incoming messages, call `future.cancel(true)` (https://github.com/eclipse-lsp4j/lsp4j/issues/770).
 
 # Extending the Protocol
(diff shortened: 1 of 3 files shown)


documentation/README.md
```diff
@@ -43,6 +43,8 @@ This will start the listening process in a new thread that reads messages from t
 
 When implementing the handlers for requests or notifications, you need to be aware that the calling thread is the thread that reads and dispatches incoming messages. Therefore, blocking it may result in reduced throughput or even a deadlock (https://github.com/eclipse-lsp4j/lsp4j/issues/775). As a general rule, message handlers should be implemented in a non-blocking, asynchronous way.
 
+For long-running work, prefer returning a `CompletableFuture` and performing the actual computation outside of the message dispatching thread. This keeps the JSON-RPC message processor responsive while the request is being handled asynchronously.
```
pisv, line 46: I think that using a code example could be the best way to clarify it, e.g. based on the actual case from #775:

````suggestion
When implementing the handlers for requests or notifications, you need to be aware that the calling thread is the thread that reads and dispatches incoming messages. Therefore, blocking it may result in reduced throughput or even a deadlock (https://github.com/eclipse-lsp4j/lsp4j/issues/775). As a general rule, message handlers should be implemented in a non-blocking, asynchronous way. For example, instead of

```java
@Override
public void didChangeWorkspaceFolders(DidChangeWorkspaceFoldersParams params) {
    build();
}
```

use

```java
@Override
public void didChangeWorkspaceFolders(DidChangeWorkspaceFoldersParams params) {
    CompletableFuture.runAsync(() -> {
        build();
    });
}
```
````

A separate example along these lines could also be added for a request handler, if necessary.

pisv (commented):
Thanks for your contribution! Returning a `CompletableFuture` as suggested in the clarification is not possible for notification methods. In general, I think that adding a code example would be the best way to clarify the non-blocking handler guidance.

pisv:
Also, @RyanBBlanck as the author of the proposed changes needs to sign [ECA](https://www.eclipse.org/legal/eca/) before this contribution can be accepted. Be sure to use the same email address when you register for the Eclipse Foundation account that you used for your commits. Thanks, and sorry for the hindrance!

